PR #1, opened 2018-04-27
Update pubspec.yaml

fmatosqg:
Enforce flutter version is >= 0.3.5

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -73,3 +73,6 @@ flutter:
   #
   # For details regarding fonts from package dependencies, 
   # see https://flutter.io/custom-fonts/#from-packages
+
+environment:
+  flutter: ">=0.3.5 <2.0.0"


pubspec.yaml
```diff
@@ -73,3 +73,6 @@ flutter:
   #
   # For details regarding fonts from package dependencies, 
   # see https://flutter.io/custom-fonts/#from-packages
+
+environment:
+  flutter: ">=0.3.5 <2.0.0"
```
dnfield: Is <2.0.0 right here?  I'm not sure what's typical for these.

fmatosqg: that's the same style from google's video player (official flutter package). I'd love to ask them why, but we can also skip the upper bound.

https://github.com/flutter/plugins/blob/master/packages/video_player/pubspec.yaml

fmatosqg: btw we can also include `sdk: ">=2.0.0"` if you want to use dart 2.0 optional keywords (`new`, `;`).

PR #6, opened 2018-05-01
Fix error on slider value > max, update flutter min version

fmatosqg:
Quick hacks I had to make to build and run,

feel free to close pull request and implement fixes in different way


dnfield:
Thanks.  I had been messing around with the default size, hadn't updated it to check whether that would cause problems.

PR #16, opened 2018-06-07
fixed centering of image in scaleCanvasToViewBox()

wkornewald:
The old code would translate the image incorrectly when xscale != yscale. Also, I've removed some code duplication.

### DIFF
--- a/lib/src/render_picture.dart
+++ b/lib/src/render_picture.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:flutter/widgets.dart';
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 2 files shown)


dnfield:
Thanks for the contribution! 

The goal of the existing code is to maintain the aspect ratio and center the image if x != y.  This code causes the image to potentially get stretched if x != y, which might be desired behavior but might not (my own preference when writing the code was to maintain aspect ratio, but it seems perfectly reasonable that someone would not want that in some use case).

@escamoteur had a similar request on slack - I think this is a good idea to make an optional parameter (something like `preserveAspectRatio`).

What do you think?

wkornewald:
My code uses the same scaling for x and y, so it should preserve the aspect ratio. It just uses the min of xscale and yscale, so it can guarantee that the image fits into the desired size.

dnfield:
Ok - I'll have time to take a closer look this weekend.

dnfield:
So your code appears to not center the image, whereas the existing code does.  I like removing the code duplication, but I'm not sure this offers a better experience.  Your code produces this result (the Dart logo wants to be rectangular, in a square box it's scaled properly but not centered):

image

Whereas the existing code produces this result (the Dart logo wants to be rectangular, it's scaled properly and centered in the square box):

image


Do you have an issue that this fixes visually?

wkornewald:
Oh yeah, I got the translate and scale in the wrong order. :) It should work now.

Side-note: Unfortunately, it's difficult to get the tests working on Windows. Minor issue: `Platform.script.path` has an incorrect leading `/`. Major issue: the PNG files still aren't exactly the same although on Linux the tests pass now.

I wrote this fix because on our app it didn't center an image in the AppBar's title correctly. So we added a BoxDecoration to the Container around the SvgPicture and got these results: https://imgur.com/a/cKYqupl

dnfield:
Ok, looks good to me now.

I'll try to look at the rendering tests on Windows - I primarily work on this stuff on a Mac.

wkornewald:
Nice, thanks. Do you have an estimate when you'll publish the next release? This fix is quite important for us.

PR #18, opened 2018-06-08
color percentages can be floats and there can be whitespace everywhere

wkornewald:
Tried following the spec more closely.

### DIFF
--- a/lib/src/svg/colors.dart
+++ b/lib/src/svg/colors.dart
@@ -32,20 +32,18 @@ Color parseColor(String colorString) {
   }
 
   // handle rgb() colors e.g. rgb(255, 255, 255)
-  if (colorString.startsWith('rgb')) {
+  if (colorString.toLowerCase().startsWith('rgb')) {
     final List<int> rgb = colorString
         .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
         .split(',')
-        .map(
-      (String rawColor) {
-        if (rawColor.endsWith('%')) {
-          final int percentage =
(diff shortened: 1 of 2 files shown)


lib/src/svg/colors.dart
```diff
@@ -32,20 +32,19 @@ Color parseColor(String colorString) {
   }
 
   // handle rgb() colors e.g. rgb(255, 255, 255)
-  if (colorString.startsWith('rgb')) {
-    final List<int> rgb = colorString
+  if (colorString.toLowerCase().startsWith('rgb')) {
+    final List<int> rgb = <int>[];
```
dnfield: I'd rather avoid using a dynamically sized array when we know the expected length.  This could be `new List<int>(4);` and then the loop below would have to be modified.  Otherwise, you could just add `rawColor = rawColor.trim()` into the function passed into `map` in the original code.

test/colors_svg_test.dart
```diff
@@ -10,7 +10,7 @@ void main() {
     expect(parseColor('#FFFFFF'), white);
     expect(parseColor('white'), white);
     expect(parseColor('rgb(255, 255, 255)'), white);
-    expect(parseColor('rgb(100%, 100%, 100%)'), white);
+    expect(parseColor('RGB(  100%   ,   100.0% ,  99.9999% )'), white);
```
dnfield: This is fine, but we might as well keep the original assertion in there as well.

wkornewald:
Awesome, thanks for the quick release.

PR #29, opened 2018-06-23
Add support for gradientTransform.

slightfoot:
@dnfield only tested on my local SVGs that needed this feature. But its simple enough that it shouldn't cause any issues. Also fixes offset bug in <linearGradient> where it previous used `bounds.left` when it should have used `bounds.top` for x, y position.

### DIFF
--- a/lib/src/svg/xml_parsers.dart
+++ b/lib/src/svg/xml_parsers.dart
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
 
   parseStops(stops, colors, offsets);
 
+  final Matrix4 transform = parseTransform(getAttribute(el, 'gradientTransform', def: null));
+  
   return (Rect bounds) {
-    final Offset from = new Offset(
-      bounds.left + (bounds.width * x1),
-      bounds.left + (bounds.height * y1),
-    );
-    final Offset to = new Offset(
-      bounds.left + (bounds.width * x2),
-      bounds.left + (bounds.height * y2),
-    );
+    final from = Vector3(bounds.left + x1, bounds.top + y1, 0.0);
+    final to = Vector3(bounds.left + x2, bounds.top + y2, 0.0);
+    if(transform != null){
+      transform.transform3(from);
+      transform.transform3(to);
+    }
 
     final Gradient gradient = new Gradient.linear(
-      from,
-      to,
+      Offset(from.x, from.y),
+      Offset(to.x, to.y),
       colors,
       offsets,
       spreadMethod,
@@ -158,6 +158,8 @@ PaintServer parseRadialGradient(XmlElement el) {
   final List<Color> colors = new List<Color>(stops.length);
   final List<double> offsets = new List<double>(stops.length);
   parseStops(stops, colors, offsets);
+
+  final Matrix4 transform = parseTransform(getAttribute(el, 'gradientTransform', def: null));
 
   return (Rect bounds) {
     final double cx = _parseDecimalOrPercentage(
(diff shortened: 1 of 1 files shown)


lib/src/svg/xml_parsers.dart
```diff
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
...
   return (Rect bounds) {
-    final Offset from = new Offset(
-      bounds.left + (bounds.width * x1),
-      bounds.left + (bounds.height * y1),
-    );
-    final Offset to = new Offset(
-      bounds.left + (bounds.width * x2),
-      bounds.left + (bounds.height * y2),
-    );
+    final from = Vector3(bounds.left + x1, bounds.top + y1, 0.0);
```
dnfield: This should be 
```
    Vector3 from = Vector3(bounds.left + (bounds.width * x1),
        bounds.top + (bounds.height * y1), 0.0);
   Vector3 to = Vector3(bounds.left + (bounds.width * x2),
        bounds.top + (bounds.height * y2), 0.0);
```

I can't remember exactly why the math needs to be that way, but without it the gradients in the samples/tests don't render properly.  You'll have to get rid of final because of something below....

lib/src/svg/xml_parsers.dart
```diff
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
...
+    final to = Vector3(bounds.left + x2, bounds.top + y2, 0.0);
+    if(transform != null){
+      transform.transform3(from);
```
dnfield: Should be `from = transform.transform3(from);`

lib/src/svg/xml_parsers.dart
```diff
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
...
+      transform.transform3(to);
```
dnfield: see line above

lib/src/svg/xml_parsers.dart
```diff
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
...
+    }
 
     final Gradient gradient = new Gradient.linear(
-      from,
-      to,
+      Offset(from.x, from.y),
```
dnfield: I haven't switched things over to avoid using `new`, so I think I'd like to keep it consistent for now and just use new on these

lib/src/svg/xml_parsers.dart
```diff
@@ -125,19 +125,19 @@ PaintServer parseLinearGradient(XmlElement el) {
...
+      Offset(to.x, to.y),
```
dnfield: `new`

dnfield:
Just for reference.. here's the before and after of a working gradient with current code/this patch currently:

image

image

lib/src/svg/xml_parsers.dart
slightfoot: they are the same, transform3 alters the parameter.

dnfield: Ahhh - I'm just used to getting bitten by similarly named methods that don't do this ;)

dnfield:
Tests are still failing, maybe because of fixing `top` there?  I'll generate new golden images for those two images though, they're not very significantly different and I'll assume they're more correct now ;)

Still would be nice to have some samples with the transform on it.

dnfield:
@slightfoot any chance you could add a couple (simple) samples using this attribute?  That will help ensure the code paths are covered properly.  I'd be happy to regenerate the failing .PNG files and add them in as well to fix the tests.

dnfield:
I believe this ended up being merged as part of #54

PR #41, opened 2018-08-04
Fix rounded rect when rx or ry is missing

gmlewis:
It turns out that if `rx` or `ry` is 0, the Dart `RRect.fromRectXY` method draws a rectangle with no rounded corners. However, Inkscape and Chrome draw the rectangle with rounded borders.

I've added a simple test case that demonstrates the problem and tried to follow a similar naming pattern to the existing rounded rectangle example.

I visually inspected the tests whose golden files changed, and I could not see any difference in the images.

Note that the original code had already made provision to duplicate the missing `rx` or `ry` value if not supplied (using `??=`), but that code was never being used because the `getAttribute` call provided a default value of '0'. Also note that I tried simply removing the `def:` parameters entirely, but it turns out that `getAttribute` then returns the empty string (which is not `null`).


gmlewis:
Just FYI... hopefully, if this PR is accepted, my next PR will be this one:
https://github.com/gmlewis/flutter_svg/commit/531ed5a03df3aea73446ad05f40f8d8b48a1158a
however, I was unable to figure out how to make a PR with a diffbase of a previous PR using GitHub like I'm used to from Critique... so I'll hold off on that PR until this one is merged to make the review easier.

dnfield:
I should have some more time to look at this more closely tomorrow - but a cursory look at it looks like it's good to me.

As far as your next one goes, I want to see support for that, but I want to avoid holding on to any XML references in the definitions - the Drawable* classes should have no knowledge of Xml* if possible.

gmlewis:
OK, I looked again, and it appears I will need your guidance on how to avoid holding onto the `XmlElement`.

gmlewis:
Here is an alternative implementation for your consideration:
https://github.com/gmlewis/flutter_svg/commit/2c51d0afe1c05b35796f2223a1403aec9ff5c011
(Edited from the first implementation, so the diffs show against that one.)

dnfield:
For reasons I don't quite understand, one of the pictures if a pixel or two different - perhaps some underlying change in Flutter?  Either way, this looks good to me.  Thanks for the contribution.

dnfield:
As for the XML part - let's open another issue or PR to discuss.  Basically, I don't want anything holding ont o XML references any longer than necessary (let Dart garbage collection pick it up ASAP).  I like the idea of being able to refer to drawing elements by ID, but let's do it without using the XML nodes any longer than necessary.

PR #45, opened 2018-08-09
Run flutter format .

gmlewis:
While working on another PR, I discovered that the master branch had not been formatted recently.


dnfield:
Let's hold off on this for now. This is almost certainly due to different flutter versions being used to format.

PR #56, opened 2018-09-07
Fix merge paint edge case

sroddy:
This fixes a subtle bug that was preventing opacity to be applied under some circumstances.

Added an asset that shows the broken behaviour


dnfield:
CI is broken on this...  But it's not related to this change.  I'll have to fix that at some point

dnfield:
Thanks for the  contribution!

PR #78, opened 2018-11-10
Skip <title> tag if in the SVG.

ncb000gt:
Rendering shouldn't fail when a title tag is provided as they are part
of the SVG spec:
https://www.w3.org/TR/SVG2/struct.html#DescriptionAndTitleElements

This skips the title tag and warns of it not being used.

Signed-off-by: Nick Campbell <nicholas.j.campbell@gmail.com>

### DIFF
--- a/lib/src/svg_parser.dart
+++ b/lib/src/svg_parser.dart
@@ -101,6 +101,9 @@ void _unhandledElement(XmlElement el, String key) {
       library: 'SVG',
       context: 'in parseSvgElement',
     ));
+  } else if (el.name.local == 'title') {
+    print("WARNING: flutter_svg doesn't support the title tag.");
+    return;
   }
   assert(() {
     if (_unhandledElements.add(el.name.local)) {


ncb000gt:
I have an SVG (which I can modify if needed, but would be a hassle) that provides a title tag. The SVGs render properly in the browser, so I thought that it would make sense to allow for "skipping" of elements if they were a part of the spec but weren't used. At the moment this PR only checks the title tag and skips, but I'm happy to expand it into something else if you think that's useful. I don't know which list of elements would fall under this description so I stuck with the `title` check for now since it directly affected me. :)

dnfield:
This library already skips title tags, and the presence of absence or one doesn't impact rendering.  It does print a warning, but it's informational only.

  Could you share the svg that isn't rendering properly?

ncb000gt:
@dnfield Ah, you're right. The issues I was seeing were related to the viewport changes you had made on master. I was seeing the warning and no SVG so thought that the issue, but I was making a bunch of other changes before I realized that having the height and width set in the SVG was causing the rendering to be off the viewable page.

Closing. Thanks!

dnfield:
Ahh.  The height/width thing is trouble.

Could you provide an example?  It may be possible to warn people if the svg will be drawn out of the viewable area, the way other widgets do.

dnfield:
Or maybe it should just throw a warning if width or height are used at all on the root element.

ncb000gt:
Yea. I am using a few svg's from https://undraw.co the specific one I had an issue with is `personal trainer` (if you search for that at https://undraw.co/search), but they all have a height and width specified on the root.

In my case, the viewport values match the height and width values. But, whenever I tried to specify the container height and width for the SVG it wasn't "filling" out the container until I removed the height and width from the root element.

That behaves differently that it appears to in a browser as I only put a height and width on a div and the SVG resized as I'd expected.

I'm not really sure how it should be handled though since I haven't done much with SVGs. I don't know what is "normal" or "expected" behavior.

PR #85, opened 2018-12-04
Parsing rewrite

### LINKED ISSUE #47: Get away from XML DOM
Ongoing effort in the parsing_rewrite branch

Long been desired to not use a DOM interface - it bloats up memory usage, creates unnecessary objects, and very easily allows for unintentional object retention.  I made some effort to avoid doing any backreferencing of XML so that a forward-only SAX or .NET style model could be used.  

This is becoming more important to me as I look at doing code generation support - the parsing logic is scattered all over right now and hard to adapt to different output models.  While taking care of that, mgiht as well bite the bullet and either use the new SAX interface in dart-xml or use a push style model (which might be a bit cleaner) that I'm working on.

dnfield:
This is close to ready.

Fixes #47 - this code should not retain any unnecessary references during parsing, and should allow for lower memory usage.


dnfield:
Green!

Still a couple minor things to shift around, and I'd like to wait to merge this until the `xml` package is available on pub instead of just a git hash.

Fixes #47

PR #92, opened 2018-12-22
Text support improvements

krispypen:
* font-weight is now parsed to flutter font-weight values
* font-family is now inherited
* dx-dy-x-y fixed, also with tspan support
* fill without stroke was broken

I also changed the example text.svg where the result before the change is:
screen shot 2018-12-22 at 13 46 56

And after (the same as in chrome and other svg viewers):
screen shot 2018-12-22 at 13 47 25

@dnfield also changed to ignore the text tests also on linux platform? because it makes no sense when you need to get a result like this https://github.com/dnfield/flutter_svg/blob/master/golden/simple/text.png


example/assets/simple/text.svg
```diff
@@ -3,14 +3,26 @@
   "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
 <svg viewBox="0 0 1000 300"
      xmlns="http://www.w3.org/2000/svg" version="1.1">
-  <desc>Example text01 - 'Hello, out there' in blue</desc>
+  <desc>Example text01</desc>
```
dnfield: Can you add this as a separate test file rather than overwriting this one?

test/golden_svg_test.dart
```diff
@@ -13,7 +13,7 @@ Iterable<File> getGoldenFileNames() sync* {
       Directory(join(root, root.endsWith('test') ? '..' : '', 'golden'));
   for (FileSystemEntity fe in dir.listSync(recursive: true)) {
     if (fe is File && fe.path.toLowerCase().endsWith('.png')) {
-      if (fe.path.toLowerCase().contains('text') && !Platform.isLinux) {
+      if (fe.path.toLowerCase().contains('text')) {
```
dnfield: Text rendering gets into some platform specific wonkiness.  These need to be run only on Linux to work correctly on CI.

tool/gen_golden.dart
```diff
@@ -41,9 +41,8 @@ Iterable<File> getSvgFileNames() sync* {
   final Directory dir = Directory('./example/assets');
   for (FileSystemEntity fe in dir.listSync(recursive: true)) {
     if (fe is File && fe.path.toLowerCase().endsWith('.svg')) {
-      // Skip text based tests unless we're on Linux - these have
-      // subtle platform specific differences.
-      if (fe.path.toLowerCase().contains('text') && !Platform.isLinux) {
+      // Skip text based tests unless we can fix the rendering on Linux
+      if (fe.path.toLowerCase().contains('text')) {
```
dnfield: Ditto this

dnfield:
This is probably fine, I'll try to do a more thorough review of the actual code change later.

krispypen:
@dnfield I tried generating a png of the new text_2.svg but even my text.png is different... so the rendering on my ubuntu 18.04.1 is different then the travis ci one.

text

compared to the one on travis:
https://github.com/dnfield/flutter_svg/blob/master/golden/simple/text.png

dnfield:
I think this looks good.  I suspect the golden difference is more faithful to Chrome at this point, so let's just replace it. I'll merge this and regenerate them in a subsequent commit.  This will require a minor version bump as well, as it's breaking for text - but in a good way.

Thanks for the contribution!  Text support is a bit of a mess right now, but this is definitely an improvement.

PR #95, opened 2018-12-31
Text improvements

dnfield:


dnfield:
@krispypen FYI - this is appyling some of what you already did to the other text anchor values, if that makes sense

PR #109, opened 2019-01-30
Should not commit pubspec.lock in publish packages

kevmoo:
https://www.dartlang.org/guides/libraries/private-files#pubspeclock


dnfield:
Thanks!

PR #113, opened 2019-02-08
Fix for SVG widget size different size depending on device

gahms:
The widget created for a SVG file should have the same point-size as the SVG viewport size. See https://github.com/gahms/flutter_svg_size_bug for demonstration of the issue.


lib/svg.dart
```diff
@@ -617,11 +617,7 @@ class _SvgPictureState extends State<SvgPicture> {
     }
 
     if (_picture != null) {
-      final double devicePixelRatio =
-          MediaQuery.of(context, nullOk: true)?.devicePixelRatio ??
-              window.devicePixelRatio;
-      final Rect viewport =
-          Offset.zero & (_picture.viewport.size * devicePixelRatio);
+      final Rect viewport = Offset.zero & _picture.viewport.size;
```
dnfield: You'll need to remov the `, window` part of the `dart:ui` import now to make analysis happy.

dnfield:
This is making a bit more sense to me now, thanks for the PR and the examples.

I'm curious about what this will do to existing clients and if/how it will break them.  I'm starting to believe that it is a mistake to use the devicePixelRatio like I am, but wondering what it will end up doing to people.

dnfield:
There are also other uses of the devicepixelRatio in render_picture.dart

gahms:
I assumed they were a bit more low level but I will have a look at the other places. With my change it renders correctly in our use case.

lib/svg.dart
gahms: Fixed. I did not realise the import warning was a result of my change.

gahms:
> There are also other uses of the devicepixelRatio in render_picture.dart

In render_picture.dart the devicepixelRatio is used for clipping. It seems to me that it should be removed from there also. In my sample project the RenderPicture ends up with a size of 27x90 and a clipping rect of 54x180 and that seems wrong.

I have very little knowledge about render objects in Flutter so the thing I am in doubt about here is if Flutter render objects work in points (logical pixels) or device (real) pixels. From what I have been able to find it is still working at points (logical pixels) level but that is just a guess.

dnfield:
Yeah the clipping should be the same size as the image.  Can we add tests for this?

gahms:
I will give it a shot.

gahms:
Why do we even have the clipping? As far as I can tell we will never paint outside the viewbox anyway?

I have added a test with a SVG I have produced myself, with a viewbox that is smaller than the image, but the produced output is clipped even if I remove line 214 in render_picture.dart (context.canvas.clipRect(viewportRect);)

dnfield:
There are a few places the clipping can go wrong.  This is a final guard.  Painting outside the intended area can have negative performance impact.

dnfield:
I may take this over - I believe it's causing serious problems when devicePixelRatio < 1.0 and it needs to be fixed for another customer ASAP

PR #118, opened 2019-02-21
Pass the PictureConfiguration to obtainKey

### LINKED ISSUE #117: SvgPicture.asset should consider the DefaultAssetBundle
For developer ergonomics, similarly to how it is passed for [Image.asset](https://docs.flutter.io/flutter/widgets/Image/Image.asset.html)

Right it now if no bundle is passed, it defaults to the rootBundle instead.

Happy to contribute a PR if this sounds like a reasonable change.

dballesteros7:
This way the ExactAssetPicture can use the DefaultAssetBundle derived
from context through the LocalPictureConfiguration instead of always defaulting
to the rootBundle. This is helpful for tests and for local widget
context assets. All other provides just ignore the configuration.

This analogous to the implementation of Image.asset in Flutter.

Fix #117

### DIFF
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -302,7 +302,7 @@ abstract class PictureProvider<T> {
     // assert(picture != null);
     final PictureStream stream = PictureStream();
     T obtainedKey;
-    obtainKey().then<void>((T key) {
+    obtainKey(picture).then<void>((T key) {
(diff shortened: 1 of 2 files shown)


dballesteros7:
This should be ready to merge, the failure was the lint failure from the other PR.

dnfield:
This is going to be released on v0.10.4

PR #119, opened 2019-02-21
Fix CI by changing linter options

dballesteros7:
Remove a deprecated one and comment out one that is triggering on new
language features.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -110,7 +110,7 @@ linter:
     # - parameter_assignments # we do this commonly
     - prefer_adjacent_string_concatenation
     - prefer_asserts_in_initializer_lists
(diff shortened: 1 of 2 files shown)


analysis_options.yaml
```diff
@@ -110,7 +110,7 @@ linter:
     # - parameter_assignments # we do this commonly
     - prefer_adjacent_string_concatenation
     - prefer_asserts_in_initializer_lists
-    - prefer_collection_literals
+    # - prefer_collection_literals # it's triggering on Set literals which are not ready https://github.com/dart-lang/linter/issues/1411
```
dnfield: Can we just add an ignore on the sets this is affecting with a TODO to change them when it's ready?

dballesteros7: Done.

PR #124, opened 2019-02-26
Update to xml 3.3.0

acherkashyn:


dnfield:
I have some work planned to do this, but I do not want to get back to the DOM based parser - I originally got away from it for a reason :)

The interface has changed.  When I first started working on it there was a bug that prevented progress, but it should be fine in 3.3.1

Is this causing problems for you right now?

acherkashyn:
@dnfield My project has a dependency on xml 3.3.0, so I can't add flutter-svg, because it causes a version clash. I didn't realize you made a decision to move away from a tree based parser. I'm curious, what was the reason for it?

dnfield:
Tree based parsing is very memory intensive, and it's not necessary for this.  I'll see if I can get my work on this into shape and open a PR for it soon.

dnfield:
Closing in favor of #125

Issue #128, opened 2019-02-28
[Suggestion] Add a precache function


Skyost:
In my app, I need to quickly display SVG pictures. As I have a loading screen, it would be great to be able to load my pictures using a `precacheImage` function (a bit like [in flutter images](https://docs.flutter.io/flutter/widgets/precacheImage.html) function).

Something like :

```dart
precacheImage(SvgPicture.asset(
      'asset.svg'
      width: 512,
      fit: BoxFit.cover,
      semanticsLabel: 'My picture',
));
```

Which returns a `Future<void>` for when it's loaded.

dnfield:
The problem is the return value of `SvgPicture.asset` is a Widget.  I think this would actually take the PictureProvider, which could make more sense.  Something like

```dart
precachePicture(ExactAssetPicture(
  svgStringDecoder,
  assetName,
   ...,
));
```

shouldn't be too hard to add though, and then if you used that asset it would be in the cache.

Skyost:
Your pull request is exactly what I was searching for ! Do you have an estimated release date for this build ?

graux:
Is there any example about how to use the precachePicture within a StatefulWidget?
I cannot load the Svg string within the initState as it cannot be async. Should I use didChangeDependencies? 
Some pointers about the steps to follow would be really helpful.
Thanks!

PR #129, opened 2019-03-01
initial implementation of precaching images

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/128

Needs tests.

@Skyost what do you think?


Skyost:
Oh yes, it seems to be great !

dnfield:
Should be soon :)  I'm going to write up a couple tests for it.  I have one other bug to get to before another release, but in the next few days it should happen.

PR #131, opened 2019-03-05
Gahms fix svg widget size

dnfield:
gahms this should complete what you started in #113.


dnfield:
Doing this as a regular merge commit so that @gahms will get credit!

PR #142, opened 2019-04-01
handle rgba colors with alpha as double

aschiavon91:
Hi buddies.

First of all, thx for this library, it's really helpful and easy to use!
I had some issues using files with objects with fill property as rgba color, so I made this little fix, hope its help in some away.

### DIFF
--- a/lib/src/svg/colors.dart
+++ b/lib/src/svg/colors.dart
@@ -34,9 +34,25 @@ Color parseColor(String colorString) {
     }
   }
 
+  // handle rgba() colors e.g. rgba(255, 255, 255, 1.0)
+  if (colorString.toLowerCase().startsWith('rgba')) {
+    final List<String> rawColorElements = colorString
+        .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
+        .split(',')
+        .map((String rawColor) => rawColor.trim())
+        .toList();
+
+    final double opacity = parseDouble(rawColorElements.removeLast());
+
+    final List<int> rgb =
+        rawColorElements.map((String rawColor) => int.parse(rawColor)).toList();
+
+    return Color.fromRGBO(rgb[0], rgb[1], rgb[2], opacity);
+  }
+
   // handle rgb() colors e.g. rgb(255, 255, 255)
   if (colorString.toLowerCase().startsWith('rgb')) {
-    final List<num> rgb = colorString
+    final List<int> rgb = colorString
         .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
         .split(',')
         .map((String rawColor) {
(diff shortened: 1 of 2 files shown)


dnfield:
Thanks for the contribution!

This is a weird case where the spec says they're supposed to be integers or percentages, but the browsers allow for doubles.  Weird.

lib/src/svg/colors.dart
```diff
@@ -36,21 +36,27 @@ Color parseColor(String colorString) {
 
   // handle rgb() colors e.g. rgb(255, 255, 255)
   if (colorString.toLowerCase().startsWith('rgb')) {
-    final List<int> rgb = colorString
+    final List<num> rgb = colorString
```
dnfield: I think we should probably keep this as a `List<int>` - things could go very wrong if we actually pass a double to the `Color` constructors that expect an `int`, not a `double`.

lib/src/svg/colors.dart
```diff
@@ -36,21 +36,27 @@ Color parseColor(String colorString) {
...
         .split(',')
         .map((String rawColor) {
       rawColor = rawColor.trim();
       if (rawColor.endsWith('%')) {
         rawColor = rawColor.substring(0, rawColor.length - 1);
         return (parseDouble(rawColor) * 2.55).round();
+      } else if (rawColor.contains('.')) {
+        return parseDouble(rawColor);
       }
       return int.parse(rawColor);
```
dnfield: It would probably make sense to just have this be `return parseDouble(rawColor).round();` instead of checking for a `.`.

lib/src/svg/colors.dart
dnfield: Oh!  Only the opacity is allowed to be a double.

lib/src/svg/colors.dart
```diff
@@ -36,21 +36,27 @@ Color parseColor(String colorString) {
...
     }).toList();
 
-    // rgba() isn't really in the spec, but Firefox supported it at one point so why not.
-    final int a = rgb.length > 3 ? rgb[3] : 255;
-    return Color.fromARGB(a, rgb[0], rgb[1], rgb[2]);
+    if (rgb.length > 3 && rgb[3] is double) {
+      return Color.fromRGBO(rgb[0], rgb[1], rgb[2], rgb[3]);
```
dnfield: We should probably just always use `RGBO`.  My test case for `rgba(255, 255, 255, 255)` isn't valid.

aschiavon91:
maybe should be better create another if condition just to rgba parse, on top of rgb condition (line 36)?

dnfield:
Sure!

aschiavon91:
What do you think about this implementation?
```
// handle rgba() colors e.g. rgba(255, 255, 255, 1.0)
  if (colorString.toLowerCase().startsWith('rgba')) {
    final List<String> rawColorElements = colorString
        .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
        .split(',')
        .map((String rawColor) => rawColor.trim())
        .toList();

    final double opacity = parseDouble(rawColorElements.removeLast());

    final List<int> rgb =
        rawColorElements.map((String rawColor) => int.parse(rawColor)).toList();

    return Color.fromRGBO(rgb[0], rgb[1], rgb[2], opacity);
  }
```

aschiavon91:
My pleasure and thanks for the help.

PR #148, opened 2019-04-04
Check transform before calling into it.

ncb000gt:
There isn't an explicit check to see if the transform was null before calling into it. This was causing a `NoSuchMethodError` trying to call `transform.storage`. Adding the check here stopped that error from occurring.

Signed-off-by: Nick Campbell <nicholas.j.campbell@gmail.com>

### DIFF
--- a/lib/src/svg/xml_parsers.dart
+++ b/lib/src/svg/xml_parsers.dart
@@ -346,16 +346,18 @@ DrawableStyle parseStyle(
     final Matrix4 transform = parseTransform(
       getAttribute(attributes, 'transform'),
     );
-    if (multiplyTransformByParent && parentStyle?.transform != null) {
-      if (transform == null) {
-        rawTransform = parentStyle.transform;
+    if (transform != null) {
+      if (multiplyTransformByParent && parentStyle?.transform != null) {
+        if (transform == null) {
+          rawTransform = parentStyle.transform;
+        } else {
+          rawTransform = Matrix4.fromFloat64List(parentStyle.transform)
+              .multiplied(transform)
+              .storage;
+        }
       } else {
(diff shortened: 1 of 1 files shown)


ncb000gt:
I'm using a few SVGs from undraw.co: here is one that had the issue, but it's not the only one (can't inline or upload svgs in github apparently):  https://42f2671d685f51e10fc6-b9fcecea3e50b3b59bdc28dead054ebc.ssl.cf5.rackcdn.com/illustrations/welcome_3gvl.svg

lib/src/svg/xml_parsers.dart
```diff
@@ -346,16 +346,18 @@ DrawableStyle parseStyle(
     final Matrix4 transform = parseTransform(
       getAttribute(attributes, 'transform'),
     );
-    if (multiplyTransformByParent && parentStyle?.transform != null) {
-      if (transform == null) {
-        rawTransform = parentStyle.transform;
+    if (transform != null) {
```
dnfield: I saw this this morning and have a different fix - I believe this would cause problems if the transform is null and you want the parent transform.

dnfield:
I was simultaneously fixing this :)  Thanks for the suggestion though!  Can you verify that my fix works now?

lib/src/svg/xml_parsers.dart
ncb000gt: Yea. I was wondering about that. I originally changed the `else` to an `else if (transform != null)`, but then decided to move it up.

Happy to nuke this pr if you have something better already or I can change it back to the `else if`. :)

dnfield: My solution was to just change `transform.storage` to `transform?.storage`.

ncb000gt:
@dnfield That appears to have fixed it! Thanks!

PR #149, opened 2019-04-06
Causes error when widget disposed

Akhilesh05:
I have scroll list, sometimes I scroll so fast that the Widget gets disposed by flutter. I undoubtedly get an error here since we are attempting to setState on a widget that is not on the screen anymore.

See error below:

> flutter: ══╡ EXCEPTION CAUGHT BY SVG ╞═══════════════════════════════════════════════════════════════════════
> flutter: The following assertion was thrown by a picture listener:
> flutter: setState() called after dispose(): _SvgPictureState#e68e1(lifecycle state: defunct, not mounted,
> flutter: stream: PictureStream#452fe(OneFramePictureStreamCompleter#f936a, Instance of 'PictureInfo', 1
> flutter: listener))
> flutter: This error happens if you call setState() on a State object for a widget that no longer appears in
> flutter: the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error
> flutter: can occur when code calls setState() from a timer or an animation callback. The preferred solution
> flutter: is to cancel the timer or stop listening to the animation in the dispose() callback. Another
> flutter: solution is to check the "mounted" property of this object before calling setState() to ensure the
> flutter: object is still in the tree.
> flutter: This error might indicate a memory leak if setState() is being called because another object is
> flutter: retaining a reference to this State object after it has been removed from the tree. To avoid memory
> flutter: leaks, consider breaking the reference to this object during dispose().
> flutter:
> flutter: When the exception was thrown, this was the stack:
> flutter: # 0      State.setState.<anonymous closure> 
> package:flutter/…/widgets/framework.dart:1095
> flutter: # 1      State.setState 
> package:flutter/…/widgets/framework.dart:1121
> flutter: # 2      _SvgPictureState._handleImageChanged 
> package:flutter_svg/svg.dart:623

### DIFF
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -620,9 +620,11 @@ class _SvgPictureState extends State<SvgPicture> {
   }
 
   void _handleImageChanged(PictureInfo imageInfo, bool synchronousCall) {
-    setState(() {
-      _picture = imageInfo;
-    });
+    if (mounted) {
+      setState(() {
+        _picture = imageInfo;
+      });
+    }
   }
 
   // Update _pictureStream to newStream, and moves the stream listener


dnfield:
Is this still reproducing for you on 0.12.3?  I introduced a different fix that should work a little more efficiently than this one.

Akhilesh05:
I don't know why the build is failing. I run flutter test and it's passing on my laptop.

I upgraded to 0.12.3
```
  flutter_svg:
    dependency: "direct main"
    description:
      name: flutter_svg
      url: "https://pub.dartlang.org"
    source: hosted
    version: "0.12.3"
```
But still same problem.
Below is my code.
```dart
  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      body: ListView.builder(
        itemCount: _leagues.length,
        itemBuilder: (BuildContext context, int index) {
          League league = _leagues[index];
          Widget image = league.logo.toString().endsWith('.svg') ? SvgPicture.network(
            league.logo.toString(),
            fit: BoxFit.scaleDown,
          ) : Image(
            image: CachedNetworkImageProvider(league.logo.toString()),
            fit: BoxFit.scaleDown,
          );
          return AspectRatio(
            aspectRatio: 3,
            child: Card(
              child: Padding(
                padding: const EdgeInsets.all(8.0),
                child: Row(
                  children: <Widget>[
                    AspectRatio(
                      aspectRatio: 1,
                      child: image,
                    ),
                    Expanded(
                      child: Center(
                        child: Text(
                          league.name,
                          style: Theme.of(context).textTheme.title,
                          overflow: TextOverflow.ellipsis,
                        ),
                      ),
                    ),
                  ],
                ),
              ),
            ),
          );
        },
      )
    );
  }
```

dnfield:
We should not need to check mounted for this..  I'll look at it.

dnfield:
I'm not able to reproduce this on master.  Can you provide a full reproduction that I can run locally?

dnfield:
Going to close this for now - please feel free to re-open if you're still reproducing this issue on the latest version of the package.

PR #158, opened 2019-05-02
Refactor error messages.

jacob314:
This refactor would be for the branch matching master.


lib/src/utilities/errors.dart
```diff
@@ -4,18 +4,20 @@ import 'package:flutter/foundation.dart';
 void reportMissingDef(String href, String methodName) {
   FlutterError.onError(
     FlutterErrorDetails(
-      exception: StateError('Failed to find definition for $href'),
```
jacob314: I could have refactored this error less but this case looks like it would be a little bit more natural as a FlutterError.
I'm also happy to keep it as it was instead of using FlutterError.

PR #175, opened 2019-06-11
[MOD]: - Agregado soporte para atributos con valor 'null'.

boanergepro:


dnfield:
This is not part of the SVG specification.  I'd rather avoid implementing things that extend the SVG spec as much as possible.

lib/src/utilities/xml.dart
```diff
@@ -49,7 +49,7 @@ String getAttribute(
     raw = _getAttribute(el, name, namespace: namespace)?.trim();
   }
 
-  return raw == '' || raw == null ? def : raw;
+  return raw == '' || raw == null || raw == 'null' ? def : raw;
```
dnfield: In this context, null could have semantic value for some reason - it might be an ID of a paint right?  This function is also used for all attribute processing in XML.  

It looks like Chrome would successfully render your SVG though.  Perhaps we can make the `parseColor` function more tolerant so that it doesn't throw?

Alternatively, we could try to make the parser more tolerant of invalid values here the way Chrome is - the problem with that is it can be difficult to tell when an invalid value should allow us to continue or not.

boanergepro: You're right, from the context of the library there is no point in the complicated subject.

PR #195, opened 2019-07-17
Add a delay by zero to make parsing async

Kavantix:
Parsing in the current version turns out to be fully synchronous and needs to be done on the main thread (it gives an exception when running in an isolate that a native method is not available).
So to fix this I added an `await Future.delayed(Duration.zero);` which makes sure that the parsing is split up into small pieces that are computed async and don't block the main thread completely.

### DIFF
--- a/lib/src/svg/parser_state.dart
+++ b/lib/src/svg/parser_state.dart
@@ -678,6 +678,8 @@ class SvgParserState {
   /// Drive the [XmlTextReader] to EOF and produce a [DrawableRoot].
   Future<DrawableRoot> parse() async {
     for (XmlEvent event in _readSubtree()) {
+      // Make sure the main thread is not completely blocked
+      await Future<void>.delayed(Duration.zero);
       if (event is XmlStartElementEvent) {
         if (startElement(event)) {
           continue;


lib/src/svg/parser_state.dart
```diff
@@ -678,6 +678,7 @@ class SvgParserState {
   /// Drive the [XmlTextReader] to EOF and produce a [DrawableRoot].
   Future<DrawableRoot> parse() async {
     for (XmlEvent event in _readSubtree()) {
+      await Future.delayed(Duration.zero);
```
dnfield: Can we just `await null` instead?

Kavantix: Apparently we can and it also works, expected that to throw an error.

lib/src/svg/parser_state.dart
```diff
@@ -678,6 +678,8 @@ class SvgParserState {
   /// Drive the [XmlTextReader] to EOF and produce a [DrawableRoot].
   Future<DrawableRoot> parse() async {
     for (XmlEvent event in _readSubtree()) {
+      // Make sure the main thread is not completely blocked
+      await null;
```
dnfield: Can you think of a way to add a test for this?  

This seems like something that could help, but I'd like to see a test that fails without this - e.g. you give it a large-ish SVG to parse and show that something else can finish before parsing is done.

dnfield: This also makes me wish we had a a performance benchmark :\

Kavantix: When I was making the unit test I found out that the behaviour only occurs in release mode.
In debug mode it already does not block the main thread so most likely the vm handles the future differently than the AOT version.
Also await by null does not work! I thought it did since the simulator did not block the main thread.
So sadly it is currently not possible to reproduce the blocking in a unit test.

The code below should however test this case if the behaviour in the unit test was the same.
``` 
  test('Parsing does not block', () async {
    const String _svg =
        '''<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg"  x="0px" y="0px"
     viewBox="0 0 500 500" style="enable-background:new 0 0 500 500;" xml:space="preserve">
      <circle id="XMLID_1_" cx="257.8" cy="226.2" r="61.5"/>
      <circle id="XMLID_3_" cx="287.3" cy="226.2" r="26.6"/>
    </svg>''';
    final List<int> _testResults = <int>[];
    Future<void> parse() async {
      await SvgParser().parse(_svg);
      _testResults.add(0);
    }

    Future<void> insert() async {
      _testResults.add(1);
    }

    final Future<void> parseFuture = parse();
    final Future<void> insertFuture = insert();

    await insertFuture;

    expect(_testResults.length, 2);
    expect(_testResults.first, 1);
  });
```

dnfield: Hmm.  I can try to think on this a bit, but maybe for now we could just add a comment in there explaining this.

Kavantix: Sorry for the late response,
For the comment do you mean explaining why the delay is needed or that a testcase cannot be written?

Also have you thought of anything that could solve the issue with the test case?

bean5:
Can't you write your own wrapper for this? Wait, isolates are usually for background stuff, right?

So you are rendering SVG in background? I'd like to understand better. I might be far newer to flutter and would like to learn more.

Kavantix:
It’s not about rendering but about parsing an svg string

Kavantix:
On newer versions of flutter/dart this seems no longer needed so I'll close this PR for now

PR #199, opened 2019-07-26
Adding mask support

krispypen:
* mask support
* added some test "golden" test files
* also fixed that parentstyles where not used with the "use" element


krispypen:
@dnfield can you first merge #200 to fix the travis build?

lib/src/vector_drawable.dart
```diff
@@ -1153,6 +1163,15 @@ class DrawableShape implements DrawableStyleable {
     } else {
       innerDraw();
     }
+    if (style.mask != null) {
+      final Paint p = Paint();
+      p.blendMode = BlendMode.dstIn;
+      p.colorFilter = ColorFilter.matrix(<double>[0,0,0,0,0, 0,0,0,0,0, 0,0,0,0,0, 0.21,0.71,0.08,0,0]); //convert grayscale values to transparency
```
dnfield: This seems right, but could you provide a link to the spec for this in a comment?

lib/src/vector_drawable.dart
```diff
@@ -1143,6 +1150,9 @@ class DrawableShape implements DrawableStyleable {
       }
     };
 
+    if (style.mask != null) {
+      canvas.saveLayer(bounds, Paint());
```
dnfield: Don't pass in the bounds here.  SaveLayer will just use the current clip anyway, and the bounds may not actually be what we want if there are transforms involved.

lib/src/vector_drawable.dart
```diff
@@ -1153,6 +1163,15 @@ class DrawableShape implements DrawableStyleable {
     } else {
       innerDraw();
     }
+    if (style.mask != null) {
+      final Paint p = Paint();
+      p.blendMode = BlendMode.dstIn;
+      p.colorFilter = ColorFilter.matrix(<double>[0,0,0,0,0, 0,0,0,0,0, 0,0,0,0,0, 0.21,0.71,0.08,0,0]); //convert grayscale values to transparency
+      canvas.saveLayer(bounds, p);
```
dnfield: Ditto on not passing in the bounds here.

lib/src/vector_drawable.dart
dnfield: Also, do we really need both of these?  Is this one required?

krispypen: You mean both invocations of saveLayer? The first one is needed to let te mask only apply on the current shape rendered in the "innerDraw" method, and the second one is needed to convert the mask to transparency and apply it to the layer

dnfield:
Thanks!

PR #200, opened 2019-07-26
Fix travis build failed

krispypen:
* "Package 'fonts-droid' has no installation candidate"
* `@virtual` is deprecated


dnfield:
Thanks!

PR #219, opened 2019-09-01
add support for hsl, hsla and transparent fill value

christianalfoni:
Thanks for an excellent library!

I am using [FramerX](https://www.framer.com/) to create SVGs. Sometimes it uses `fill="transparent"`and `fill="hsl(0,0%,0%)"` etc. I have been manually converting this, though thought it would be nice to just get it out of the box as there are probably other scenarios where this is an issue as well.

The HSL(A) conversion logic is taken from: https://github.com/MichaelFenwick/Color

So big shoutout to @MichaelFenwick for also creating an excellent package ❤️

### DIFF
--- a/lib/src/svg/colors.dart
+++ b/lib/src/svg/colors.dart
@@ -52,28 +52,27 @@ Color parseColor(String colorString) {
 
   // Conversion code from: https://github.com/MichaelFenwick/Color, thanks :)
   if (colorString.toLowerCase().startsWith('hsl')) {
-    List<num> rgb = <num>[0, 0 , 0];
     final List<int> values = colorString
         .substring(colorString.indexOf('(') + 1, colorString.indexOf(')'))
         .split(',')
         .map((String rawColor) {
-          rawColor = rawColor.trim();
-          
-          if (rawColor.endsWith('%')) {
-            rawColor = rawColor.substring(0, rawColor.length - 1);
-          }
+      rawColor = rawColor.trim();
 
-          if (rawColor.contains('.')) {
-            return (parseDouble(rawColor) * 2.55).round();
-          }
+      if (rawColor.endsWith('%')) {
+        rawColor = rawColor.substring(0, rawColor.length - 1);
+      }
 
-          return int.parse(rawColor);
-        }).toList();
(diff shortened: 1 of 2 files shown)


christianalfoni:
Related to coverage I guess I could test a bunch more HSL colors. But yeah, let me know

lib/src/svg/colors.dart
```diff
@@ -50,19 +50,77 @@ Color parseColor(String colorString) {
     return Color.fromRGBO(rgb[0], rgb[1], rgb[2], opacity);
   }
 
+  // Conversion code from: https://github.com/MichaelFenwick/Color, thanks :)
+  if (colorString.toLowerCase().startsWith('hsl')) {
+    List<num> rgb = <num>[0, 0 , 0];
```
dnfield: Please use either `int` or `double` here (avoid `num`) and elsewhere, as appropriate.

Also, pleae `dartfmt` this code.

lib/src/svg/colors.dart
```diff
@@ -50,19 +50,77 @@ Color parseColor(String colorString) {
...
+
+    final num hue = values[0] / 360 % 1;
+    final num saturation = values[1] / 100;
+    final num luminance = values[2] / 100;
+    final num alpha = values.length > 3 ? values[3] : 255;
+
+    if (hue < 1 / 6) {
+      rgb[0] = 1;
+      rgb[1] = hue * 6;
+    } else if (hue < 2 / 6) {
```
dnfield: Please add some tests for these branches. It should be pretty trivial :)

lib/src/svg/colors.dart
dnfield: Also, please move the declaration of this down closer to where it is used.

christianalfoni:
Okay, I believe things are in order now :)

christianalfoni:
btw, do you have an ETA on next release? Do not mean to seem impatient or anything. No rush for me. Just wondering :-)

dnfield:
I'll shoot to do one today

dnfield:
Published as v0.14.1

PR #228, opened 2019-09-23
Only fix image size on one dimension

### LINKED ISSUE #227: Image sizes should only be fixed on one dimension
Consider the following repro, which I have attached in full as a zip at the end of this issue:

```dart
import 'package:flutter/material.dart';
import 'package:flutter_svg/flutter_svg.dart';

void main() => runApp(MyApp());

class MyApp extends StatelessWidget {
  @override
    final column = Column(
      children: <Widget>[
        Row(children: [
          ImageWithText(),
        ]),
        Row(
          children: <Widget>[
            ImageWithText(),
            ImageWithText(),
          ],
        ),
        Row(
          children: <Widget>[
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
          ],
        ),
        Row(
          children: <Widget>[
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
          ],
        ),
        Row(
          children: <Widget>[
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
            ImageWithText(),
          ],
        ),
      ],
    );
    final scaffold = Scaffold(
      appBar: AppBar(),
      body: column,
    );
    final materialApp = MaterialApp(
      home: scaffold,
    );
    return materialApp;
  }
}

class ImageWithText extends StatelessWidget {
  @override
    final image = SvgPicture.asset('assets/sample.svg');
    //final image = Image.asset('assets/bitmap.png');
    final imageContainer = Container(
      child: image,
      color: Colors.amber,
    );
    final text = Text('Hello');
    final column = Column(
      children: <Widget>[
        imageContainer,
        text,
      ],
    );
    final expanded = Expanded(
      child: column,
    );
    return expanded;
  }
}
```

When displaying a bitmap (via `Image.asset`) I see:

Screenshot_20190923-165832

But when displaying an SVG (via `SvgPicture.asset`) I see:

Screenshot_20190923-165851

Notice how as the horizontal space gets tighter, the images correctly get narrower, but the SVGs incorrectly retain their height as they get narrower.

Debugging through the code, I believe the problem is [here](https://github.com/dnfield/flutter_svg/blob/master/lib/svg.dart#L712-L713). Setting both the width and height means that the `SizedBox` only shrinks in one direction. I think it should be setting whichever dimension is largest. I tried commenting out the assignment to `height` (which works for this case) and it worked as expected. I think the code should be something like:

```dart
if (width == null && height == null) {
    width = viewport.width > viewport.height ? viewport.width : null;
    height = viewport.height > viewport.width ? viewport.height : null;
} 
```

[repro_svg_sizing_issue.zip](https://github.com/dnfield/flutter_svg/files/3640939/repro_svg_sizing_issue.zip)

kentcb:
Fixes #227

This is intended to fix #227, but I may have missed something. The outer `SizedBox` was the culprit and in my opinion shouldn't be there. This PR removes it and my testing has shown it to work correctly now.

### DIFF
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -709,8 +709,8 @@ class _SvgPictureState extends State<SvgPicture> {
       double width = widget.width;
       double height = widget.height;
       if (width == null && height == null) {
-        width = viewport.width;
-        height = viewport.height;
+        width = viewport.width > viewport.height ? viewport.width : null;
+        height = viewport.height > viewport.width ? viewport.height : null;
       } else if (height != null) {
         width = height / viewport.height * viewport.width;
       } else if (width != null) {
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -704,33 +704,16 @@ class _SvgPictureState extends State<SvgPicture> {
     }
 
     if (_picture != null) {
-      final Rect viewport = Offset.zero & _picture.viewport.size;
-
-      double width = widget.width;
-      double height = widget.height;
-      if (width == null && height == null) {
-        width = viewport.width > viewport.height ? viewport.width : null;
-        height = viewport.height > viewport.width ? viewport.height : null;
-      } else if (height != null) {
-        width = height / viewport.height * viewport.width;
-      } else if (width != null) {
-        height = width / viewport.width * viewport.height;
-      }
-
       return _maybeWrapWithSemantics(
-        SizedBox(
-          width: width,
-          height: height,
-          child: FittedBox(
-            fit: widget.fit,
-            alignment: widget.alignment,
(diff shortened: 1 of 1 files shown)


kentcb:
Actually, this fix doesn't work. Still looking into this - please don't merge.

dnfield:
This is breaking.  We have to come up with something else here.  See discussion at linked issue.

PR #230, opened 2019-09-30
DRAFT: Initial web support draft

### LINKED ISSUE #173: Support for Web
Hi

I'm using flutter_svg in my apps, but I need to use in a simple web page that will be created with Flutter Web.

Does flutter_svg supports Flutter for Web or it will support soon? This is in your plans?

jonasbark:
Would fix #173 

This is more of a draft than a fully working pull request. Things I had to do:
- change HTTP Get method to use `http` package that is available on web platform as well
- import universal_io so we can still expose SvgPicture.file
- store SVG xml string in `PictureInfo` so we can use it for HtmlPlatformView

Some problems exist:
- in order to have correct scaling on the Web you'd be required to add a width + height to the SvgPicture constructor. I don't think there's a way around it.
- having a separate PictureInfo class for the Web would be smarter
- code repetition in svg.dart, svg_web.dart and svg_io.dart
- colorFilter and some other parameters are ignored on the web
- on the web we shouldn't need to download a network image SVG first before displaying it


dnfield:
Thanks for doing this!

I'm a bit concerned with some of the approach here.  It seems to me that we shouldn't need to take dependencies on new packages for this and it would be better to avoid that. For instance, for a network picture we shouldn't use our own httpclient, we should just set that as the parameter on the foreign element we're injecting anyway.

I'm also concerned with how breaking this is.  We also should be able to still support color filtering on the web (it shoul djust be a matter of setting the right CSS property/ies on the element).

WDYT?

jonasbark:
I also would like to avoid adding dependencies but as far as I can tell it's the only way to keep exposing SvgPicture.file apart from two other breaking possibilities:
- we could change the constructor of SvgPicture.file to accept the files' path as String instead.
- we could expose an additional import (e.g. flutter_svg_io.dart), that exposes a way to use SvgPicture with a File.

I'll check for color filtering.

It's fine if this PR cannot be merged for a while - maybe it's just a bit too early for this :-)

dnfield:
We should probably just make .file throw an unsupported error on web

jonasbark:
That already happens in universal_io.dart.
Even if we would do that right from the constructor in svg_web.dart, then we'd still need to import `dart:io` in svg_web.dart - which is not allowed.

MaximilianKlein:
I tried your changes (with some minor changes for my use-case) and they worked for me using assets, thanks :) .

I also found a proposal for federated plugins in Flutter (https://docs.google.com/document/d/1LD7QjmzJZLCopUrFAAE98wOUQpjmguyGTN2wd_89Srs/edit#heading=h.n1m70qye8n4g) that might solve the issue with the imports when it lands. I tried it a bit in a fork (but there is no tooling support from Flutter yet).
With federated plugins you can separate the implementation from the actual exported interface and pub selects the correct platform package for you. flutter_svg could host web and mobile implementations here or someone else could provide a web-implementation that you might endorse or not. Seems quite interesting.

stefanrusek:
@jonasbark  One way to work around requiring both width and height.. the root svg xml tag has attributes, specifically viewbox and sometimes height and width, so the default size and aspect ratio can easily be parsed out.

TannerYoung:
Hi there, I'm willing to look into the colorFilter problem if it's something in this codebase and not upstream. Let me know if there's a way I can help and thanks for the hard work!

TannerYoung:
Also it doesn't look like alignment is properly working. I'm guessing that has to do with the width + height?

bean5:
I hope this works and that it gets accepted!

I'm fine with specifying size. Small price to pay for this!

bean5:
I hope this works and that it gets accepted!

I'm fine with specifying size. Small price to pay for this!

wanjm:
is it possible just to <svg> of html instead of draw svg itself;

wanjm:
someone told me image.network can use asserts svg path as parameter in web;

fotiDim:
Shouldn't this be merged? Works fine for me so far.

dnfield:
This patch is out of date and there are alternative approaches being taken to achieve the same effect. Flutter SVG already has support on the CanvasKit backend, and should soon have support on the HTML backend.

PR #238, opened 2019-10-16
fix image scaling

### LINKED ISSUE #229: Incorrect image size when setting both width and height in conjunction with BoxFit.contain
Setting both width and height in conjunction with BoxFit.contain leads to incorrect image dimensions. 

Example code:

```
SvgPicture.asset(
  'some_image.svg',
  height: 64,
  width: 64,
  fit: BoxFit.contain,
)
```

According to the documentation of [BoxFit.contain](https://api.flutter.dev/flutter/painting/BoxFit-class.html), the box should be sized as large as possible while still containing the source entirely within the target dimensions. However, if the width of some_image.svg is larger than 64 pixels, the resulting SvgPicture grows beyond its specified dimensions of 64x64.

It seems like this is caused by the following code in svg.dart:

```
double width = widget.width;
double height = widget.height;
if (width == null && height == null) {
  width = viewport.width;
  height = viewport.height;
} else if (height != null) {
  width = height / viewport.height * viewport.width; <- width is overwritten 
} else if (width != null) {
  height = width / viewport.width * viewport.height; <- height is overwritten 
}
```

There should be no viewport calculations whatsoever if both width and height are given.

fl0cke:
This PR fixes #229.

### DIFF
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -711,10 +711,10 @@ class _SvgPictureState extends State<SvgPicture> {
       if (width == null && height == null) {
         width = viewport.width;
         height = viewport.height;
-      } else if (height != null) {
-        width = height / viewport.height * viewport.width;
-      } else if (width != null) {
+      } else if (height == null) {
         height = width / viewport.width * viewport.height;
+      } else if (width == null) {
+        width = height / viewport.height * viewport.width;
       }
 
       return _maybeWrapWithSemantics(


lazingor:
I don't think the proposed code changes any of the logic (and I don't know that there's a bug here). 
`height == null` and `width != null` are equivalent in the `else if`, as are `width == null` and `height != null` in the other `else if`.

fl0cke:
When both `width` and `height` are set, the original code executes the second if branch and overwrites `width` with its own value. This should not happen. My code fixed that by inverting the null checks so that none of the branches are taken in said case.

lazingor:
Ah, I see.

dnfield:
This has a linting issue .

I'm also surprised it doesn't break any tests.  Can we add a test that covers this?

gilbertoalexsantos:
@fl0cke Can you fix the linting issue?

Just replace the last ``` width == null ``` with ```??=```

It should be like this:
```
else if (height == null) {
    height = width / viewport.width * viewport.height;
} else {
    width ??= height / viewport.height * viewport.width;
}
```

Parkar99:
It's been more than six months. Any update on this?

dnfield:
I'm going to close this - if someone wants to take it over in a new PR that's fine, or if the original author is interested in adding a test that would work too (we can reopen).

PR #266, opened 2019-12-18
Update goldens

chingjun:
https://github.com/flutter/engine/pull/13777 changed the rendering,
update the goldens to fix the tests


chingjun:
Note that the engine commit hasn't been rolled into the framework tree yet, this PR is going to unblock the roll.

chingjun:
We're going to revert the change in the engine tree instead, closing this PR.

PR #277, opened 2020-01-19
Adding an error widget

PoloLacoste:


lib/svg.dart
```diff
@@ -582,6 +588,9 @@ class SvgPicture extends StatefulWidget {
   /// The placeholder to use while fetching, decoding, and parsing the SVG data.
   final WidgetBuilder placeholderBuilder;
 
+  /// The errorBuilder to use when there is an error.
```
dnfield: Nit - maybe just `The widget to display if there is an error`?

lib/svg.dart
```diff
@@ -692,6 +701,9 @@ class _SvgPictureState extends State<SvgPicture> {
   @override
   Widget build(BuildContext context) {
     Widget _maybeWrapWithSemantics(Widget child) {
+      if(_picture == null && widget.errorBuilder != null){
+        return widget.errorBuilder(context);
+      }
```
dnfield: This will never get called.  It's already wrapped in an `if (_picture != null)`.

lib/src/picture_stream.dart
```diff
@@ -296,6 +296,7 @@ class OneFramePictureStreamCompleter extends PictureStreamCompleter {
       {InformationCollector informationCollector})
       : assert(picture != null) {
     picture.then<void>(setPicture, onError: (dynamic error, StackTrace stack) {
+	  setPicture(null);
```
dnfield: I don't think this is enough to distinguish between a picture that has an error and a picture that is still loading (because of an ongoing network fetch for example).

We probably need to pipe some callback through to here for errors.

lib/svg.dart
```diff
@@ -582,6 +588,9 @@ class SvgPicture extends StatefulWidget {
   /// The placeholder to use while fetching, decoding, and parsing the SVG data.
   final WidgetBuilder placeholderBuilder;
 
+  /// The widget to display if there is an error.
+  final Widget errorWidget;
```
dnfield: I think this should be a WidgetBuilder, right?

lib/src/picture_stream.dart
```diff
@@ -237,7 +238,7 @@ abstract class PictureStreamCompleter extends Diagnosticable {
         List<_PictureListenerPair>.from(_listeners);
     for (_PictureListenerPair listenerPair in localListeners) {
       try {
-        listenerPair.listener(picture, false);
+        listenerPair.listener(picture, false, _error);
```
dnfield: I think if this throws the error builder won't get called, right?

PoloLacoste:
```dart
void _handleImageChanged(PictureInfo imageInfo, bool synchronousCall, bool error) {
    setState(() {
      _picture = imageInfo;
	  _error = error;
    });
  }
```

It's catch in the function above (in the `_SvgPictureState` class).
I don't know where the error builder should be call apart in `_SvgPictureState`.

dnfield: If you look at line 243, there's a check for an error listener.  We should be bubbling that back up as well.

shliama:
So what's happening?

dnfield:
The contributor appears to have lost interest in continuing with this PR. If someone else is interested in picking it up that would be fine, keeping in mind whatever review was left from here.

PR #299, opened 2020-02-24
Diagnosticable => DiagnosticableMixin

gspencergoog:
Converting these to DiagnosticableMixin, in preparation for eventually deprecating `Diagnosticable`.

### DIFF
--- a/lib/src/picture_stream.dart
+++ b/lib/src/picture_stream.dart
@@ -85,7 +85,7 @@ typedef PictureListener = Function(PictureInfo image, bool synchronousCall);
 ///
 ///  * [PictureProvider], which has an example that includes the use of an
 ///    [PictureStream] in a [Widget].
-class PictureStream extends Diagnosticable {
+class PictureStream with DiagnosticableMixin {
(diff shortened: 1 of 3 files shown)


lib/src/picture_stream.dart
```diff
@@ -85,7 +85,7 @@ typedef PictureListener = Function(PictureInfo image, bool synchronousCall);
 ///
 ///  * [PictureProvider], which has an example that includes the use of an
 ///    [PictureStream] in a [Widget].
-class PictureStream extends Diagnosticable {
+class PictureStream with DiagnosticableMixin {
```
dnfield: We should bump the minimum Flutter versino to v1.6.7 for this, which is where it was introduced AFAICT

gspencergoog: Good point. Added that to this PR, and added a comment in the CHANGELOG.  Should I also bump the version number, or will you do that when you publish?

dnfield: Yeah I can take care of that, thanks!

PR #306, opened 2020-03-09
Don't default to black when no colour is defined.

### LINKED ISSUE #296: <use> refering to an element with a fill is not carrying over the fill
https://raw.githubusercontent.com/hjnilsson/country-flags/master/svg/cn.svg

The yellow five stars turned black.

pento:
Fixes #296.

When parsing an SVG, the `parseColor()` function returns `colorBlack` if no colour is defined. While this behaviour is correct when rendering the SVG, it causes problems when merging colours. #296 is a good example of this, where `_Elements.use()` calls `parseStyle()`, but there's no way for it to know if the `parent.style.fill.color` is black because it's defined as black, or black because it wasn't defined, and `parseColor()` defaulted to black.

This PR included a test render to show that the changed works.

The change to `use_opacity_grid.png` appears to be correct, when comparing the render to how `use_opacity_grid.svg` appears in Chrome.


dnfield:
Published this and your last change in v0.17.3.  Thanks again

PR #308, opened 2020-03-09
Add an opacity param to DrawablePaint

### LINKED ISSUE #307: Regression: a shape element with no color some opacity is rendering as solid black now
E.g.

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 264 264">
  <rect width="264" height="264" fill="#e8f0fe"/>
  <path d="M184.32,121.12H79.38v8H184.32Z" opacity="0.1" style="isolation:isolate"/>
</svg>
```

No longer renders correctly after #306 

Unfortunately there was no test covering this.

/cc @pento

pento:
Fixes #307.

After #306, we still need to store the opacity when the colour isn't defined.

This PR adds an `opacity` param to the `DrawablePaint` class, and applies it to the `color` when the `DrawablePaint` is transformed to a `Paint` object.


pento:
@dnfield: Given that this PR is causing the `blend_and_mask` test to change, I'm not 100% sure I'm applying the opacity consistently, everywhere.

I think it's pretty close, though, I can dig further sometime later.

pento:
Oh, this causes changes to the `text` test files, too.

I suspect that `DrawablePaint` needs to keep track of the parent element's `DrawablePaint`, which will allow it to retrieve the parent `color` in `toFlutterPaint()`, rather than defaulting to `colorBlack`. It should only default to `colorBlack` when there's no parent (ie, in the `<svg>` element`).

dnfield:
I _just_ pushed up a commit that I believe fixes it - see a3515a3

dnfield:
Well, at least the commit I just added restores the old behavior for shapes without losing the fix you created.

If there's a more spec compliant way of doing this I'm all for it.  For now I wanted to squash the regression even if we're not 100% correct.

pento:
Noice. 🙂 I'll rebase this PR and play around with it a bit later, thanks for taking care of the regression.

pento:
I poked at it a bit more, I think the direction I was heading was over-engineering it. I'll close this PR. 🙂

PR #331, opened 2020-04-20
Fix issue with the SvgPicture widget sizing when no width and height

ramsestom:
should fix https://github.com/dnfield/flutter_svg/issues/227

### DIFF
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -703,37 +703,45 @@ class _SvgPictureState extends State<SvgPicture> {
       );
     }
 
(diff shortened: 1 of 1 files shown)


dnfield:
Sorry it took me so long to get to this - I meant to reply and then lost track of it. I think this makes sense, but it now has a merge conflict and lacks a test. I'm fixing up the merge conflicts and adding a test from the issue you filed. However, I don't seem to have permissions to push to your branch. I'm going to just open a new PR for this with your commit as part of it.

dnfield:
Landed in #500. Sorry for the delay.

PR #339, opened 2020-05-05
Upgrade xml dependency

Zazo032:
Fixes https://github.com/dnfield/flutter_svg/issues/320


Zazo032:
@dnfield a version bump to use latest `xml` version, all tests are passing locally.

Zazo032:
Not really, seems like they were automatically added to migrate to latest iOS templates. Should I revert them?

dnfield:
Ok, no worries.  Thanks for doing this!

kpsroka:
When I try to run `flutter pub get` on a freshly cloned repo with the newest stable (1.17.0), I get a depedency resolution error:
```
➜  flutter_svg git:(master) flutter pub get
Because every version of flutter_test from sdk depends on xml 3.6.1 and flutter_svg depends on xml ^4.1.0, flutter_test from sdk is forbidden.
So, because flutter_svg depends on flutter_test any from sdk, version solving failed.
Running "flutter pub get" in flutter_svg...                             
pub get failed (1; So, because flutter_svg depends on flutter_test any from sdk, version solving failed.)
```

dnfield:
What version of flutter?

kpsroka:
Today's new stable (v1.17.0).

dnfield:
Hmm. We'll probably have to bump the minimum constraint on master for this then.  Unfortunately, there's a clash between packages here and it's not compatible.

dnfield:
Yeah, https://github.com/flutter/flutter/pull/54144 fixes this issue. You can try patching that into Stable, it may just work.  Otherwise, it's available as of 1.18.0-6.0.pre. I've bumped the minimum constraint on master to this version.

Issue #317, opened 2020-03-28
SvgPicture.network() not passing through headers


javabada:
Hey, looks like http headers aren't being passed through to the actual request.

Not sure how to unit test this however.

https://github.com/dnfield/flutter_svg/blob/35970c251899cf366084e9bb8eeea24a48b4fbe8/lib/src/picture_provider.dart#L507

daniellampl:
The fix is still not in version 0.19.0

dnfield:
Fixed as of v0.20.0-nullsafety.2

PR #342, opened 2020-05-07
Fix Issue #317 headers were not passed to http client.

m-oehme:
Using SvgPicture.network with headers was not possible since headers were not passed to the http get request function

fixes https://github.com/dnfield/flutter_svg/issues/317


dnfield:
I really thought this had been fixed before. 

Can you please add a test covering this?

stijnie2210:
I'm currently running into this issue. Is there any chance this can be merged soon? @dnfield

dnfield:
It needs a test

stijnie2210:
Is there any chance you could pick this up? This seems like a pretty
serious bug since headers are now just useless in the network constructor

dnfield:
Added a test for this and some version bumps.

I realize this would be nice to backport into a stable supported version - that should be doable  if absolutely necessary, but for now I'm putting it into the nullsafety version which requires a pretty recent version of Flutter.

dnfield:
Published as v0.20.0-nullsafety.2

PR #354, opened 2020-05-29
await runZoned Futures in the tests

mehmetf:
With an upcoming change to widget tester, this will cause failures. See: https://github.com/flutter/flutter/pull/57894


dnfield:
Ci failures are unrelated to this change. Will fix those separately

PR #356, opened 2020-05-31
Updating dependencies

luanpotter:
I got a conflict in my project because another dependency was using the latest `xml` version (`4.2.0`). Since it's always good to keep everything up to date, I took the liberty to update all dependencies (they required either patch or minor upgrades, so should be safe). I also ran the whole build pipeline (format, analysis & tests) and it seems to be fine.

I would greatly appreciate if a new version could be released with the updated xml dependency :)

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,16 +5,16 @@ version: 0.17.4
 
 dependencies:
   path_drawing: ^0.4.1
-  xml: ^4.1.0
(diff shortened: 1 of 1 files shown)


luanpotter:
@dnfield sorry to tag you, but any chance of getting this deployed? lmk if there is anything else I can help with

dnfield:
Thanks!

PR #371, opened 2020-06-29
Close http client when request finishes

spiritinlife:
I think we should close the httpClient here, since we are not reusing the http client.


spiritinlife:
I am not sure about this flutter's `NetworkImage` does not seem to close the client either but they do seem keep a static httpclient to be reused 
`static final HttpClient _sharedHttpClient = HttpClient()..autoUncompress = false;` 

Maybe this library should do something similar.

Issue #330, opened 2020-04-14
SVG with multiple gradient transforms throwing error


VikasJilla:
```
<radialGradient cx="0%" cy="0%" fx="0%" fy="0%" r="99.0351718%" gradientTransform="translate(0.000000,0.000000),scale(0.891892,1.000000),rotate(51.447275),translate(-0.000000,-0.000000)" id="radialGradient-20">
 <stop stop-color="#FBAA30" stop-opacity="0.511117788" offset="0%"></stop>
 <stop stop-color="#FBC230" stop-opacity="0" offset="100%"></stop>
</radialGradient>
```

I/flutter (28075): Log.DEBUG : =================== CAUGHT FLUTTER ERROR
I/flutter (28075): Log.ERROR : Bad state: Unsupported transform: ,translate
I/flutter (28075): Stack Trace: #0      parseTransform (package:flutter_svg/src/svg/parsers.dart:109:7)
I/flutter (28075): #1      _Elements.radialGradient (package:flutter_svg/src/svg/parser_state.dart:244:39)
I/flutter (28075): #2      SvgParserState.parse (package:flutter_svg/src/svg/parser_state.dart:776:26)
I/flutter (28075): <asynchronous suspension>
I/flutter (28075): #3      SvgParser.parse (package:flutter_svg/parser.dart:14:60)
I/flutter (28075): #4      Svg.fromSvgString (package:flutter_svg/svg.dart:110:25)
I/flutter (28075): #5      Svg.svgPictureStringDecoder (package:flutter_svg/svg.dart:72:36)
I/flutter (28075): #6      SvgPicture.svgStringDecoder.<anonymous closure> (package:flutter_svg/svg.dart:531:15)
I/flutter (28075): #7      AssetBundlePictureProvider._loadAsync (package:flutter_svg/src/picture_provider.dart:457:19)
I/flutter (28075): <asynchronous suspension>
I/flutter (28075): #8      AssetBundlePictureProvider.load (package:flutter_svg/src/picture_provider.dart:435:43)
I/flutter (28075): #9      PictureProvider.resolve.<anonymous closure>.<anonymous closure> (package:flutter_svg/src/picture_provider.dart:327:17)
I/flutter (28075): #10     PictureCache.putIfAbsent (package:flutter_svg/src/picture_cache.dart:67:22)
I/flutter (28075): #11     Pict
[...]

PR #378, opened 2020-07-08
Allow multiple transforms in an attribute 

muzzah:
Regular Expression was including commas in the matching group. Added a small check to exclude any commas from matches and added tests to validate.


muzzah:
This aims to fix #330

PR #381, opened 2020-07-11
Expose SVG group and path 'id' attributes to DrawableRoot

dr953:
Exposing SVG group/path 'id' is essential to be able to identify and access specific groups/paths from a SVG file, at runtime.

At the most basic level, it allows for a simple inclusion/exclusion of specific groups/paths, before drawing to a canvas.
It can also allow for a runtime SVG modification (color, stroke, etc.).

There are many applications for such functionalities.


dnfield:
Support for this just landed in another PR.

PR #392, opened 2020-08-05
support id

zl910627:
id supports is needed

### DIFF
--- a/lib/src/vector_drawable.dart
+++ b/lib/src/vector_drawable.dart
@@ -23,6 +23,8 @@ final Paint _grayscaleDstInPaint = Paint()
 /// Base interface for vector drawing.
 @immutable
 abstract class Drawable {
+  String get id;
+
   /// Whether this [Drawable] would be visible if [draw]n.
   bool get hasDrawableContent;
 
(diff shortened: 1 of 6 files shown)


lib/src/vector_drawable.dart
```diff
@@ -23,6 +23,8 @@ final Paint _grayscaleDstInPaint = Paint()
 /// Base interface for vector drawing.
 @immutable
 abstract class Drawable {
+  String get id;
```
dnfield, line 26: Please add a docstring.

Probably something like

> A string that should uniquely identify this [Drawable] within its [DrawableRoot].

Is fine.

lib/src/vector_drawable.dart
```diff
@@ -859,6 +865,9 @@ class DrawableRoot implements DrawableParent {
   @override
   final List<Drawable> children;
 
+  @override
+  String get id => 'root';
```
dnfield, line 869: I think we should avoid this - someone might use that string in an SVG, and we should just take the `@id` attribute from the `<svg>` element itself instead.

lib/src/vector_drawable.dart
zl910627: Just added document as you suggested, thank you

zl910627:
Hello @dnfield 
I have updated PR, and please check.
Looking forward updated package on Pub.
Thank you
Zheng

lib/src/vector_drawable.dart
zl910627: Yes, this has been fixed on new commit.
I also have added test case for this.
Thank you

Issue #384, opened 2020-07-17
SvgPicture should expose clipBehavior


liyuqian:
SvgPicture now exposes the fit parameter from FittedBox, but not clipBehavior.

The clipBehavior will be needed to properly clip the content. See https://github.com/flutter/flutter/issues/61755

Specifically, some Google apps (e.g., customer dream and customer money) use SvgPicture and they'll need the clipBehavior so we can roll the clip change into Google.

liyuqian:
CC @dnfield

dnfield:
Are you able to open a patch that exposes it? That should be fine.

liyuqian:
Sure, I can write a patch. I guess that I'll also need to add some tests to this repo. Is there a CI test that I should pay attention to?

dnfield:
You can probably add something to https://github.com/dnfield/flutter_svg/blob/master/test/widget_svg_test.dart

PR #402, opened 2020-08-18
Expose clipBehavior from FittedBox

liyuqian:
Fixes https://github.com/dnfield/flutter_svg/issues/384


liyuqian:
@dnfield

liyuqian:
@dnfield thanks for the quick review! Do we have to bump some version number so I can make Google internal projects use this change?

dnfield:
This should roll into google as is - it pulls from git. I'll take care of bumping the version when I prep a new pub release.

PR #422, opened 2020-09-23
Remove unused dart:async imports

MichaelRFairhurst:
As of Dart 2.1, Future and Stream are exported by dart:core


dnfield (approved):
LGTM, thanks

PR #452, opened 2020-11-19
Allow options around caching for color filter(s)

dnfield:
Some use cases are seeing a performance decrease for not caching the color filters. This allow clients to migrate back or forth more easily, especially since these changes cause golden file testing changes.

b/173063280


dnfield:
CI failure is related to coverage format - I'll look at that separately. Landing this.

PR #465, opened 2020-12-17
Ignore dead code warning in preparation for flow analysis improvement.

stereotype441:
When https://github.com/dart-lang/language/issues/1274 (Infer
non-nullability from local boolean variables) is implemented, flow
analysis will detect that code like this is unreachable:

    bool hasArguments = false;
    ...
    if (hasArguments) {
      // Not reachable
    }

To prepare for this improvement, we need to add an "ignore" comment to
ignore the dead_code warning that will be generated.

### DIFF
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -159,9 +159,6 @@ class PictureConfiguration {
     result.write('PictureConfiguration(');
     bool hasArguments = false;
     if (bundle != null) {
-      if (hasArguments) { // ignore: dead_code
-        result.write(', ');
-      }
       result.write('bundle: $bundle');
       hasArguments = true;
     }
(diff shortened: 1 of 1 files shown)


lib/src/picture_provider.dart
```diff
@@ -159,7 +159,7 @@ class PictureConfiguration {
     result.write('PictureConfiguration(');
     bool hasArguments = false;
     if (bundle != null) {
-      if (hasArguments) {
+      if (hasArguments) { // ignore: dead_code
```
dnfield, line 162: But this isn't dead. Ok the subsequent iteration it's used.

dnfield (approved):
My mistake this was probably a copy paste error or something I mistakenly thought was in a loop

dnfield:
Ci failure is the coveralls issue which I really need to fix at some point.  Tests are passing.

PR #474, opened 2021-01-17
Add warningsAsErrors option for parsing SVGs

### LINKED ISSUE #473: Detect unhandled elements
@dnfield is it possible in a dry-run to detect if all parts in a SVGs are supported yet? 🤔
The idea is to prevent uploads of broken SVG files in an internal tool built on Flutter.

What I was thinking about is using the [`svgPictureStringDecoder`](https://pub.dev/documentation/flutter_svg/latest/svg/Svg/svgPictureStringDecoder.html) and then detecting if any errors occur.

Looking at the source code I've found the [`_unhandledElement` variable](https://github.com/dnfield/flutter_svg/blob/f2a180642019f503cbe4b542c1b89e74fd584227/lib/src/svg/parser_state.dart#L17), which is filled, but never accessed.
It's currently, only used for throwing errors once per entry, but the library doesn't do anything else with it. (We would also have to track unhandled XML properties)

How hard would it be to report these kinds of warnings as part of the PictureInfo?
Also, would it also work for the web? I guess so, because we only parse the SVG, but not render it yet?
https://github.com/dnfield/flutter_svg/issues/173#issuecomment-503901756

IchordeDionysos:
## Description

Adds a dry run mode which will always throw an error aborting the rendering process.

The dry run can be used to check if an SVG is supported by the library.

## Related Issues

Closes #473

### DIFF
--- a/lib/parser.dart
+++ b/lib/parser.dart
@@ -8,8 +8,18 @@ class SvgParser {
   /// Parses SVG from a string to a [DrawableRoot].
   ///
   /// The [key] parameter is used for debugging purposes.
-  Future<DrawableRoot> parse(String str, {String? key}) async {
-    final SvgParserState state = SvgParserState(xml.parseEvents(str), key);
+  ///
+  /// The [dryRun] detects if an SVG contains contains unsupported features.
+  /// If true the function will throw with an error.
+  /// If false only warnings are logged to the console.
+  /// Defaults to false.
+  Future<DrawableRoot> parse(
+    String str, {
+    String? key,
+    bool dryRun = false,
+  }) async {
+    final SvgParserState state =
+        SvgParserState(xml.parseEvents(str), key, dryRun);
     return await state.parse();
   }
 }
--- a/lib/parser.dart
+++ b/lib/parser.dart
@@ -9,17 +9,19 @@ class SvgParser {
   ///
   /// The [key] parameter is used for debugging purposes.
   ///
-  /// The [dryRun] detects if an SVG contains contains unsupported features.
-  /// If true the function will throw with an error.
-  /// If false only warnings are logged to the console.
+  /// By default SVG parsing will only log warnings when detecting unsupported
+  /// elements in an SVG.
+  /// If [warningsAsErrors] is true the function will throw with an error
+  /// instead.
+  /// You might want to set this to true for test and to false at runtime.
   /// Defaults to false.
   Future<DrawableRoot> parse(
     String str, {
     String? key,
-    bool dryRun = false,
+    bool warningsAsErrors = false,
   }) async {
     final SvgParserState state =
-        SvgParserState(xml.parseEvents(str), key, dryRun);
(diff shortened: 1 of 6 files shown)


IchordeDionysos:
@dnfield I've implemented a dry-run that does exactly what you've meant in #473

I've run the tests locally and they worked for me.
Should I care about the tests failing because one is about the Dart version and the second about an unrelated assert.

lib/src/svg/parser_state.dart
```diff
@@ -698,12 +707,14 @@ class _SvgGroupTuple {
 /// Maintains state while pushing an [XmlPushReader] through the SVG tree.
 class SvgParserState {
   /// Creates a new [SvgParserState].
-  SvgParserState(Iterable<XmlEvent> events, this._key)
-      : assert(events != null), // ignore: unnecessary_null_comparison
+  SvgParserState(Iterable<XmlEvent> events, this._key, this._dryRun)
+      : assert(events != null),
+        // ignore: unnecessary_null_comparison
```
dnfield, line 712: This needs to be moved up to the line above or the one above that.

dnfield: swapping these two will keep dartfmt happy.

lib/parser.dart
```diff
@@ -8,8 +8,18 @@ class SvgParser {
...
-    final SvgParserState state = SvgParserState(xml.parseEvents(str), key);
+  ///
+  /// The [dryRun] detects if an SVG contains contains unsupported features.
+  /// If true the function will throw with an error.
+  /// If false only warnings are logged to the console.
+  /// Defaults to false.
+  Future<DrawableRoot> parse(
+    String str, {
+    String? key,
+    bool dryRun = false,
```
dnfield, line 19: Here and elsewhere: I wonder if we can come up with a better name than `dryRun`. Seeing this pop up in my IDE as a suggestion doesn't immediately suggest anything to me.

Maybe something like `bool warningsAsErrors`, or `bool unhandledElementsAsErrors`?

dnfield (commented):
This overall looks pretty good. The `master` CI should pass for this, but `stable` won't for other reasons related to NNBD not being available on stable yet.

I'd also like to see if we can come up with a better name than `dryRun`. We might also suggest that users set it to true in tests even if they have it as false at runtime.

lib/parser.dart
IchordeDionysos: Makes sense to me:
I would go with `warningsAsErrors` as in the future the library might also warn about unsupported SVG attributes which is currently missing, but might be helpful.

IchordeDionysos:
@dnfield Well, I have no luck :D
Now coveralls doesn't want to run.

I've adjusted the PR with your suggestions!

dnfield:
That intermittently breaks and I haven't gotten to the bottom of it :\

dnfield (approved):

PR #477, opened 2021-01-22
Remove return stmts from Future<void>.catchError

srawlins:
A similar, but oddly reversed, fix to https://github.com/dnfield/flutter_svg/commit/272f9929f009e0b4e1413095ab4d2e3f3233b8d8.

The issue here is that a `Future<void>`'s catchError onError handler should return `void`. I think this fix is a no-op, because the `Future<PictureStream>` was immediately dropped on the floor. I the important bit in there (`onError(exception, stack);`) remains.

The dart analyzer will start reporting code like this when https://github.com/dart-lang/sdk/issues/35825 is closed.

### DIFF
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -340,6 +340,7 @@ abstract class PictureProvider<T> {
     ).catchError((Object exception, StackTrace stack) async {
       if (onError != null) {
         onError(exception, stack);
+        return;
       }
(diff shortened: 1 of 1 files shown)


dnfield (approved):

dnfield:
I need to fix CI in this repo, but I'm waiting for Flutter to roll stable which will make it a bit easier for me.

PR #489, opened 2021-02-08
Added Picture Info callback

dmk-rib:
Could you create a new preview package with these changes


dnfield:
Can you revert the changes to example for this?

I'm not sure why you'd use this rather than working directly with the picture providers and adding a listener. You could still pass the same provider into the SvgPicture widget.

dmk-rib:
@dnfield, yea you are right. But the Offset value from SVG is still missing. I've made a new [pull request](https://github.com/dnfield/flutter_svg/pull/501) only with this small update. Could you take a look [there](https://github.com/dnfield/flutter_svg/pull/501)?

PR #494, opened 2021-02-21
Add an alternative tool that converts SVG to custom paint

amorenew:
Some developers would need to control the SVG code instead of loading it from the asset folder.
Like changing colors, or anything else to make a cool animation to the SVG image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,3 +196,7 @@ licensed under the Apache license.
 Please submit SVGs this can't render properly (e.g. that don't render here the
 way they do in chrome), as long as they're not using anything "probably out of
 scope" (above).
+
+## Alternative tool to convert SVG to a custom paint code
+- You could use custom paint to make animation or any effect you want as you have full control on the SVG code
+[Flutter Shape Maker](https://fluttershapemaker.com)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -196,3 +196,7 @@ licensed under the Apache license.
 Please submit SVGs this can't render properly (e.g. that don't render here the
 way they do in chrome), as long as they're not using anything "probably out of
 scope" (above).
+
+## Alternative tool to convert SVG to a custom paint code
+- You could use custom paint to make animation or any effect you want as you have full control on the SVG code
```
dnfield, line 201: I'm not really sure I agree with how this gives "full control" any more than the flutter_svg library does. In either case you can modify your SVG source, or you can manipulate objects that they produce (e.g. the `Drawable` object output of flutter_svg, or the custom paint code of the Flutter Shape Maker).

I'm happy to just make this change myself, but WDYT of just having a section here that just lists alternative vector graphics libraries, and links to things like `Flutter Shape Maker`, `Rive`, etc.? I'd rather avoid appearing to endorse any of these libraries for any particular feature they may or may not have, and just say what they do and how to get more information.

amorenew: @dnfield So you mean I should keep the title but removing the description?

amorenew: @dnfield I updated my PR and I just mentioned a simple description about it

Issue #493, opened 2021-02-21
SVGs to Flutter Custom Paint Code


amorenew:
I found the following tool to convert SVG to custom paint code
I share it so it could be an alternative
https://fluttershapemaker.com/

dnfield:
Cool. I'd be happy to add this to the readme as an alternative if the author would like.

amorenew:
@dnfield I created a PR
https://github.com/dnfield/flutter_svg/pull/494

dnfield:
Mentioned this in the PR I just opened, but on a couple SVGs I just tried this tool produces pretty questionable output. I'm also a bit concerned about the design model here - having a web service rather than e.g. a codegen solution is probably harder to deal with to make sure your code stays in sync with what the designer gives you. I'm still fine with linking it though.

dnfield:
The code it produces will also have trouble because it ends up doing lots of work in the paint phase and doesn't cache anything.

amorenew:
@dnfield Author mentioned that he will make static variables to improve the performance.
Still, the tool is not perfect but I think it is a cool idea

amorenew:
@dnfield I approved your PR, Send me the Dash bird :)

PR #495, opened 2021-02-21
Add alternative implementations

dnfield:
Alternative for https://github.com/dnfield/flutter_svg/pull/494

Fixes https://github.com/dnfield/flutter_svg/issues/493

@amorenew

I've tried a couple seemingly simple SVGs on this tool though and it produces ... questionable output. I can see some that it does pretty correctly, but other seemingly simple ones that just seem dead wrong to me.


amorenew (approved):
Cool LGTM

PR #501, opened 2021-02-27
Extended PictureInfo with offset

dmk-rib:


taras-mashlyanka:
Any news on this? We are using several other packages which depends on flutter_svg and it would really help us if this PR will be merged.

PR #506, opened 2021-03-06
Improved instruction to install to make it clearer

inaps:
That was not too clear that I have to add assets to pubspec, so I've updated README file to make user experience smoother

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,28 @@ do not break known-good renderings.
 
 Basic usage (to create an SVG rendering widget from an asset):
 
+1. Install dependecies in pubspec.yaml:
 ```dart
-final String assetName = 'assets/image.svg';
+dependencies:
+  flutter:
+    sdk: flutter
+  flutter_svg:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,8 +26,28 @@ do not break known-good renderings.
...
+  flutter:
+    sdk: flutter
+  flutter_svg:
+```
+2. Create `assets` folder on the root level
+3. Add some svg to assets. We've added `icon-product.svg` for example
+4. Inject assets in pubspec.yaml
+```dart
+flutter:
+  uses-material-design: true
```
PiN73, line 41: This line is not needed I think

README.md
```diff
@@ -26,8 +26,28 @@ do not break known-good renderings.
 
 Basic usage (to create an SVG rendering widget from an asset):
 
+1. Install dependecies in pubspec.yaml:
 ```dart
```
PiN73, line 30: Here and below it'd be better to use `  ```yaml` instead of `  ```dart` for proper code coloring

PR #525, opened 2021-03-31
Refactor attribute localName check

meiwin:
Existing implementation uses replaceFirst API which will create a new String. This change replaces it with `localName` API of XmlEventAttribute which internally uses substring.

The end result is overall faster parsing.


dnfield (commented):
LGTM

dnfield:
Failure is a flake in coveralls :\

PR #538, opened 2021-04-22
expose cache on PictureProvider

dnfield:
Otherwise, users cannot change the default maximum size of the cache from 1000 (!).


dnfield:
Coverage decrease is coverage for deprecated lines. Silly. Merging.

PR #546, opened 2021-05-05
Improve performance of SVG parsing

jefflim-google:
The previous code used to iterate through a list of XmlEventAttributes for every call to getAttribute. The newer code constructs a Map, and looks up the attribute name directly.

This seems to show a 30% overall reduction to svg parsing costs.

### DIFF
--- a/lib/src/svg/parser_state.dart
+++ b/lib/src/svg/parser_state.dart
@@ -750,7 +750,7 @@ class SvgParserState {
         depth -= 1;
         assert(depth >= 0);
       }
-      _currentAttributes = {};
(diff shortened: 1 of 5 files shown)


dnfield (commented):
The style in this repo is more like Flutter than pedantic/google internal. Can you take a look at the analysis issues? Otherwise LGTM

jefflim-google:
Fixed all lint issues. PTAL

jefflim-google:
Remaining issue is with coveralls, but I'm not too sure what I should do for that.

dnfield:
I'll fix up coveralls separately. LGTM.

PR #556, opened 2021-05-22
Implemented text-style parser

### LINKED ISSUE #517: text font-style="italic" does not work
Most text attributes work, but not italic. This fails:
    `<text` x="20" y="35" font-style="italic"  `font-family="serif">text</text>`
but this works:
    `<text` x="20" y="35" font-weight="bold"  `font-family="serif">text</text>`

beroso:
This PR fixes #517 .


beroso:
@dnfield I updated the text_2.svg sample but I'm having a hard time to generate the golden for this file.
That's because I'm working on MacOS platform, and the gen_golden skips text based tests unless it runs on Linux.
Could you help me to update this golden to pass the checks?

example/assets/simple/text_2.svg
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" standalone="no"?>
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" 
   "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg viewBox="0 0 1000 300"
+<svg viewBox="0 0 1000 350"
```
dnfield, line 4: Rather than modifying this test, could you just create a new SVG with the feature(s) you want? That way there's no confusion about whether this could have broken a previous golden.

Once you upload it, I can regenerate the goldens on Linux and upload them (or you can, if you have access to a Linux machine). I use Linux becuase it's generally available on CI...

beroso: Got it! I will make it today

dnfield (approved):
LGTM, thanks!

guman24:
Why this is not available in current library (branch). I am still unable to use this feature. Always it's not rendering font-style=italic

PR #562, opened 2021-05-27
Add text-decoration support

jnschulze:
This PR adds support for `text-decoration`, `text-decoration-style` and `text-decoration-color`.

This doesn't include a golden file for `text_5.svg` yet as `gen_golden.dart` excludes `text_*.svg` on Windows machines.


dnfield (approved):
Great, LGTM! Added the golden.

PR #563, opened 2021-05-27
Add support for xml:space

jnschulze:
Handle the `xml:space` attribute for `text` and `tspan` elements.

If set to `preserve`, only tabs and newlines will be removed.


dnfield (commented):
Nice!

Can you add a test for this?

jnschulze:
I've added a simple test case, however, it just relies on a text element being in the tree or not (as empty elements get removed unless having `xml:space` set to `preserve`).

Is it possible to get the actual text back from `DrawableText`?

dnfield (approved):
LGTM, thanks

PR #583, opened 2021-08-06
add a simple scrolling benchmark for the example

flar:


flar:
I have no idea why the pub get step and analyzer fail in CI, they work fine on my machine.

dnfield:
.cirrus.yml is missing a pub get :\

flar:
Thanks Dan! Now the analyzer is complaining about one of the SVG classes' composition. I've never seen the `Avoid defining a class that contains only static members` warning before. Also notable is the fact that the class in question seems to have been "all-static" for some time and the warning has been explicitly enabled for a few years now.

dnfield:
Going to land this, will fix the lint issue separately.

PR #584, opened 2021-08-11
Unused parameter `bounds` of method draw()

mishkov:
I think programmers should be aware that this parameter is not used

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,8 @@ svgRoot.scaleCanvasToViewBox(canvas);
 // Optional, but probably normally desireable: ensure the SVG isn't rendered
 // outside of the viewbox bounds
 svgRoot.clipCanvasToViewBox(canvas);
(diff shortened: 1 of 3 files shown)


mishkov:
> I'm not a fan of the underscore notation either

Sorry, I was wrong [the underscore is only used for callbacks](https://dart.dev/guides/language/effective-dart/style#:~:text=This%20guideline%20is,it%20isn%E2%80%99t%20used)

mishkov:
> This parameter may be used but is not guaranteed.

Yes, but it does not affect the execution result

> I'm not a fan of the underscore notation either

Sorry, I was wrong [the underscore is only used for callbacks](https://dart.dev/guides/language/effective-dart/style#:~:text=This%20guideline%20is,it%20isn%E2%80%99t%20used)

> Perhaps we should just update the docs?

Yes it will be great

PR #588, opened 2021-08-21
Add hacky support for currentColor

### LINKED ISSUE #31: Svg won't render when color specified as 'currentColor'.
Hi - the attached svg has been downloaded from https://feathericons.com/. It specifies the stroke as `stroke="currentColor"`, which apparently is a kind of web standard for the inherited color. (?)

If I remove both `fill` and `stroke` definitions, the svg is rendered with `fill` as whatever is passed in as the `color` field; but the stroke is neither defined nor used.

Would it be possible to get `stroke` to honour the `color` directive from the widget's config? Or (better still for me) `colors.dart` to recognise and correctly return the `currentColor` wherever it's used?

[home.svg.zip](https://github.com/dnfield/flutter_svg/files/2150917/home.svg.zip)

lstanden:
I didn't want to modify a whole bunch of SVG files by hand, so I added a quick hacky solution to support `currentColor`.  It just sets it to black, which should work.  Maybe.
Would resolve #31


dnfield (changes requested):
This doesn't really work - currentColor should be looking at the color of the parent element, if any. Black is probably fine if it's on the root.

dnfield:
One idea: it might be wroth adding a `currentColor` property to the widget and using it on the root if the root asks for currentColor.

lstanden:
Yup.  This works for my specific use case right now, but I might try and find some time next week to update it to account for a parent color.

xepozz:
@lstanden could you please finish work?

lstanden:
@dnfield I had some time to work on this today, and applied a different method that allows the color to be passed down.  It seems this was already effectively supported if you had an empty `fill` or `stroke` attribute.  There was a couple of places for stroke (in particular) that didn't enjoy `parseColor` returning null, but that was easily fixed.

example/test_driver/bench_test.dart
```diff
@@ -14,7 +14,8 @@ Future<void> main() async {
     final Timeline timeline = await driver.traceAction(() async {
       await driver.scroll(
         view,
-        0, -3400,
+        0,
```
lstanden, line 17: I guess `flutter format .` caught this, since I didn't touch it.

lib/src/avd/xml_parsers.dart
```diff
@@ -70,7 +70,7 @@ Matrix4 parseTransform(List<XmlAttribute> el) {
 /// Parses an AVD stroke related attributes to a [DrawablePaint].
 DrawablePaint? parseStroke(List<XmlAttribute> el, Rect bounds) {
   final String? rawStroke =
-      getAttribute(el, 'strokeColor', def: null, namespace: androidNS);
+      getAttribute(el, 'stroke', def: null, namespace: androidNS);
```
lstanden, line 73: I think this was always a bug?  `strokeColor` is not a valid XML attribute for SVG as far as I can tell?

dnfield: This is for AVDs, which are different.

example/assets/svg_currentcolor.svg
```diff
@@ -0,0 +1,7 @@
+<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+    <g fill="green" stroke="green">
+        <rect width="50" height="50" fill="currentcolor" />
+        <circle r="25" cx="70" cy="70" stroke="currentcolor" fill="none" stroke-width="5" />
+        <circle r="25" cx="90" cy="90" stroke="pink" fill="none" stroke-width="3" />
+    </g>
+</svg>
```
dnfield, line 7: nit: newline at EOF

example/test_driver/bench_test.dart
dnfield: It changes sometimes depending on the flutter version, no worries.

lib/src/avd/xml_parsers.dart
dnfield: (so please revert this file :)

lib/src/svg/xml_parsers.dart
```diff
@@ -217,7 +217,8 @@ DrawablePaint? parseStroke(
     PaintingStyle.stroke,
     color: rawStroke == ''
         ? (parentStroke?.color ?? colorBlack).withOpacity(opacity)
-        : parseColor(rawStroke)!.withOpacity(opacity),
+        : (parseColor(rawStroke) ?? parentStroke?.color ?? colorBlack)
+            .withOpacity(opacity),
```
dnfield, line 221: This still doesn't seem quite right to me. 

I think we want a new property on `SvgPicture` called `currentColor` that will set a default current color. 

Then, we'd want `parseColor` to take an optional `currentColor` parameter, which is initially set to that value, and subsequently set to `parentStroke.color`.

How does that sound to you?

lstanden: I don't know if that's really necessary?  The whole idea is to inherit the color from the parent element if you use `currentColor`, which is exactly what the code currently does if you use `fill=""`.  The change to stroke was specifically to match the behavior of the fill handling, which already did this logic.

dnfield: Yes, but the default for `stroke` is `none`, which I think this will not handle quite right...

dnfield: Or rather, if `none` is explicitly specified.

lstanden: `parseColor` already return `null` for the `none` color.  The code previously failed when `parseColor` returns a null value.  I'm trying to understand the change in behavior here.

lstanden: `parseStroke` also already has an escape clause for this:
```
  if (rawStroke == 'none') {
    return DrawablePaint.empty;
  }
 ```

I guess a test case that fails would help me to understand what's wrong here... do you have a contrived example?

dnfield: Ahh ok. In that case I think this is ok. We could go back and add a property for `currentColor` separately.

dnfield (approved):

PR #592, opened 2021-08-27
Improve performance when an animation is present over the SVG

dnfield:
- Avoids calls to `drawPicture`, which is bad for current raster cache implementation in the engine
- Makes sure we reuse the `Picture` we create as much as possible, and dispose of it when done
- Adds a benchmark with the ghostscript tiger having a progress indicator animating on top of it.


dnfield:
Note to self: land this as a separate version.

dnfield:
Pointed out by @goderbauer - this patch wasn't setting the `isComplexHint` on the added PictureLayer, meaning it was less likely to get raster cached.

Issue #558, opened 2021-05-24
Publish stable 1.x version


danilofuchs:
Currently, this package only provides alpha versions, such as 0.22.0

This means every minor change (0.23.0, ...) are interpreted as breaking changes by Dart Pub.

Other dependencies might depend on a specific version of flutter_svg, so it blocks some straightforward upgrades.

I believe this package has already proven its stability and production readyness. (We've been using it extensively for 1 year without a single bug).

This is a minor incovenience when using the library for a long period. It is totally manageable, but is happens every time a new version is published.

Anyways, thank you for this amazing piece of software!

dnfield:
Yeah it's probably time. There are more things I was hoping to get done, and I'm still likely to err on the side of caution when bumping versions (sometimes people are broken because of golden tests for example).

dnfield:
So the two things I'd like to have before a 1.0 are:

- Out of Order defs (#102)
- Filter effects (#53)

danilofuchs:
Nice!
I myself missed support for filters recently and had to use PNGs instead
I understand these issues are complex to solve, and I'm afraid few people are qualified enough to act on this domain. Perhaps another team member from @flutter could help push these forward?

It would be nice to break down these in smaller deliveries (create an interface for rendering filters, and implement them bit by bit) so people can help doing small PRs and we all have partial support sooner (partial support is OK before 1.0 IMO).

dnfield:
Yeah. I have a WIP patch to support out of order defs - it's something I work on when I have cycles :) The filters might be helped by that as well. Part of my concern has been that filters, as specified, are really hard to do performantly. SVG in general isn't a great interface for performance :\

dnfield:
This is done. I still don't have out of order defs or filters, but it's ok.

Issue #193, opened 2019-07-10
Update to latest stable release


Eddydpyl:
Flutter was recently updated in its stable channel, with breaking changes for this library.

dnfield:
Use v0.13.0 and it should work.  

I'm gii g to leave this one open because I'm getting reports of this frequently.  There was a breaking change this stable that breaks 0.12.x and lower, but 0.13.x and up will work.

Eddydpyl:
Can confirm that v0.13.0 works as expected. Thanks!

dnfield:
I think most people are probably past this by now - will close and re-open if people keep filing bugs about it again :)

guilhermeagostinelli:
Not working with 1.17.0 (latest Flutter SDK release on stable channel). @dnfield Could you re-open this?

dnfield:
That's a new bug then. Please file a new issue with instructions on how to reproduce

guilhermeagostinelli:
Ok, here it is: #341

Issue #166, opened 2019-05-22
My colored SVG is rendered as black?


RegisSaffi:
Hey, i have SVG images and they are colored, when i import one in my flutter app,it come in black color,while i did not set the color. what is the problem?

MeshkaniMohammad:
Same problem observed!

Tolga-GA:
Why can't we just see the default svg color? I can see that my svg file is being brushed with the primary color.

AresVampire:
I have same problem

dnfield:
Can you provide a sample SVG that demonstrates this?

AresVampire:
[Sample.txt](https://github.com/dnfield/flutter_svg/files/3213267/Sample.txt)

This is the file. I have renamed the extension to .txt for uploading.

dnfield:
@AresVampire your problem is using a style element which isn't supported - see #105

toanalien:
Try this: https://github.com/dnfield/flutter_svg/issues/105#issuecomment-479724077

thealiflab:
You can watch this answer from StackOverflow

https://stackoverflow.com/a/62748066/11681472

agersea:
Same problem, my SVG assets render black when using SvgPicture.asset.  I do not believe they are corrupted as they display properly on my Mac and in Android Studio.

FoolishRobo:
[flutter_svg](https://pub.dev/packages/flutter_svg) can not understand [Internal or Embedded CSS](https://www.geeksforgeeks.org/types-of-css-cascading-style-sheet/).

Hence you need to convert the internal css `<style>.a{fill:#566e83;}</style>` into inline css.

Example:-
Convert this `<path class="a" ...` to `<path style="fill:#566e83;" ...`

daixianceng:
resolved: I use webview to display svg:

```dart
import 'package:webview_flutter/webview_flutter.dart';

Container(
  height: 100,
  width: 200,
  child: WebView(
    initialUrl: 'about:blank',
    onWebViewCreated:
        (WebViewController webViewController) async {
      String svgText =
          await rootBundle.loadString('assets/white-logo.svg');
      String fileText =
          '<html><style>html,body {background: #C2185B;padding: 0;margin: 0;} svg {width: 100vw; height: 100vh;}</style><body>' +
              svgText +
              '</body></html>';
      webViewController.loadUrl(
        Uri.dataFromString(
          fileText,
          mimeType: 'text/html',
[...]

Issue #144, opened 2019-04-02
SVG Network Error handling?


skybur:
How do I handle error when loading SVG from network? Currently I only use `SvgPicture.network`, when there is no internet connection the widget throws exception and the image won't reload even if I use `setState()`.
Thanks in advance

Qwin:
+1

dnfield:
I suspect there's a bug where the cache is getting a bad entry if the network is down.  I don't suppose you have a test case for this? :)

Qwin:
Well my main problem is that the error that gets thrown cannot be caught by a try / catch as it happens asynchronously. Is there a way to catch it and then retry manually ?

dnfield:
So `SvgPicture.network` is really just a convenience wrapper. Error handling for it is kind of tough and could probably be improved, but I strongly suspect you'd have a better time just handling the network retrieval separately and using `SvgPicture.string` on the downloaded content.

skybur:
I forgot I made this issue. Actually I already solved this by using `precachePicture`. There I caught the exception so the widget can automatically reload when changing page/rebuilding.

Qwin:
@dnfield Thats my current workaround sadly, it would be great to see future updates where the error can be caught using a callback handler / delegate that you pass to the SvgPicture.

Nolence:
Couldn't figure out that `precachePicture` thing. Just in case anyone gets stuck on this, here's the way I did it

```dart
FutureBuilder(
  future: http.get(iconUrl),
  builder: (context, snapshot) {
    if (snapshot.data == null)
      return CircularProgressIndicator();

    final response = snapshot.data as http.Response;

    if (![200, 201].contains(response.statusCode) ||
        snapshot.hasError) return Placeholder();

    final fileString = response.body;

    return SvgPicture.string(fileString);
  },
),
```

PR #602, opened 2021-09-22
Add support for `currentColor` to `SvgPicture`

bselwe:
Adds support for [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) to `SvgPicture`.

- Introduced `SvgTheme` used when parsing an SVG picture (includes `currentColor`).
- Introduced `DefaultSvgTheme`, an inherited widget used to provide `SvgTheme` to `SvgPicture`.
- The `SvgPicture` provides the `currentColor` (from the widget's property if given, otherwise fallbacks to the `currentColor` inherited from `DefaultSvgTheme`) to the `SvgParserState`.
- `SvgParserState` now parses the [`color`](https://www.w3.org/TR/SVG11/color.html) attribute of SVG elements and provides the value to descendant elements. If a nested element defines its color as `currentColor`, then it uses the latest (deepest) `color` defined on its ancestor element. If no `color` is defined on ancestors, then it fallbacks to the value of `currentColor` inherited from `SvgPicture`.
- Refactored `AvdPicture` not to extend `SvgPicture`.

### DIFF
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -323,6 +323,8 @@ abstract class PictureProvider<T> {
   final ColorFilter? colorFilter;
 
   /// The default color applied to SVG elements that inherit the color property.
+  @visibleForTesting
+  Color? get currentColor => _currentColor;
   Color? _currentColor;
 
   /// Sets the [_currentColor] to [color].
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -201,6 +201,24 @@ class PictureConfiguration {
   }
 }
 
+/// A type of the [PictureProvider].
+enum PictureProviderType {
+  /// Refers to the provider of type [ExactAssetPicture].
+  asset,
+
+  /// Refers to the provider of type [NetworkPicture].
+  network,
+
+  /// Refers to the provider of type [FilePicture].
+  file,
+
+  /// Refers to the provider of type [MemoryPicture].
+  memory,
+
+  /// Refers to the provider of type [StringPicture].
+  string,
+}
+
 /// Identifies a picture without committing to the precise final asset. This
 /// allows a set of pictures to be identified and for the precise picture to later
 /// be resolved based on the environment, e.g. the device pixel ratio.
--- a/lib/src/picture_provider.dart
+++ b/lib/src/picture_provider.dart
@@ -447,19 +447,20 @@ abstract class AssetBundlePictureProvider
   /// const constructors so that they can be used in const expressions.
   AssetBundlePictureProvider(this.decoderBuilder, ColorFilter? colorFilter)
       : assert(decoderBuilder != null), // ignore: unnecessary_null_comparison
-        _decoder = decoderBuilder(null),
+        decoder = decoderBuilder(null),
         super(colorFilter);
 
-  /// The decoder builder to build a [_decoder] when [currentColor] changes.
+  /// The decoder builder to build a [decoder] when [currentColor] changes.
   final PictureInfoDecoderBuilder<String> decoderBuilder;
 
   /// The decoder to use to turn a string into a [PictureInfo] object.
-  PictureInfoDecoder<String> _decoder;
+  @visibleForTesting
+  PictureInfoDecoder<String> decoder;
 
   @override
   set currentColor(Color? color) {
     _currentColor = color;
-    _decoder = decoderBuilder(_currentColor);
+    decoder = decoderBuilder(_currentColor);
   }
 
   /// Converts a key into an [PictureStreamCompleter], and begins fetching the
@@ -483,7 +484,7 @@ abstract class AssetBundlePictureProvider
       AssetBundlePictureKey key, PictureErrorListener? onError) async {
     final String data = await key.bundle.loadString(key.name);
     if (onError != null) {
-      return _decoder(
+      return decoder(
         data,
         key.colorFilter,
         key.toString(),
@@ -492,7 +493,7 @@ abstract class AssetBundlePictureProvider
         return Future<PictureInfo>.error(error, stack);
       });
     }
-    return _decoder(data, key.colorFilter, key.toString());
+    return decoder(data, key.colorFilter, key.toString());
   }
 }
 
@@ -513,14 +514,15 @@ class NetworkPicture extends PictureProvider<NetworkPicture> {
   NetworkPicture(this.decoderBuilder, this.url,
       {this.headers, ColorFilter? colorFilter})
       : assert(url != null), // ignore: unnecessary_null_comparison
-        _decoder = decoderBuilder(null),
+        decoder = decoderBuilder(null),
         super(colorFilter);
 
-  /// The decoder builder to build a [_decoder] when [currentColor] changes.
+  /// The decoder builder to build a [decoder] when [currentColor] changes.
   final PictureInfoDecoderBuilder<Uint8List> decoderBuilder;
 
   /// The decoder to use to turn a [Uint8List] into a [PictureInfo] object.
-  PictureInfoDecoder<Uint8List> _decoder;
+  @visibleForTesting
+  PictureInfoDecoder<Uint8List> decoder;
 
   /// The URL from which the picture will be fetched.
   final String url;
@@ -531,7 +533,7 @@ class NetworkPicture extends PictureProvider<NetworkPicture> {
   @override
   set currentColor(Color? color) {
     _currentColor = color;
-    _decoder = decoderBuilder(_currentColor);
+    decoder = decoderBuilder(_currentColor);
   }
 
   @override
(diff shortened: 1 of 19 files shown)


lib/src/svg/parser_state.dart
```diff
@@ -148,6 +151,7 @@ class _Elements {
         parserState._definitions,
         parserState.rootBounds,
         parent.style,
+        currentColor: parserState.theme?.currentColor,
```
dnfield, line 154: This doesn't seem right to me. The currentColor should be inherited at this point.

dnfield: IOW, the theme should only apply it to the root, not to children.

dnfield: (here and below)

lib/src/svg/theme.dart
```diff
@@ -0,0 +1,30 @@
...
+/// A theme used when decoding an SVG picture.
+class SvgTheme {
+  /// Instantiates an SVG theme with the [colorFilter] and [currentColor].
+  const SvgTheme({
+    this.colorFilter,
+    this.currentColor,
+  });
+
+  /// The color filter, if any, to apply to this widget.
+  final ColorFilter? colorFilter;
```
dnfield, line 12: This seems confusing to add here - it's not used quite the same way as `currentColor`, and I'd rather not encourage people to use `colorFilter` since it's kind of expensive.

lib/src/svg/theme.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:ui';
+
+/// A theme used when decoding an SVG picture.
+class SvgTheme {
+  /// Instantiates an SVG theme with the [colorFilter] and [currentColor].
+  const SvgTheme({
+    this.colorFilter,
+    this.currentColor,
```
dnfield, line 8: It looks like this is present here and as an individual property on the widget. 

I think I would have expected this to be an inherited widget, or perhaps for there to be a `DefaultSvgTheme` widget that's an inherited widget that the child widget(s) can use if one isn't specified.

lib/src/svg/parser_state.dart
bselwe: Thanks. I am thinking of two approaches for that: 
1. Add `currentColor` to `DrawableParent`. Populate the root SVG element `DrawableRoot` with `parserState.theme?.currentColor`, populate all the following SVG group elements `DrawableGroup` with `parent.currentColor`. In this way, this proeprty would be available at each drawable through `parent.currentColor`.
2. The same as above, but instead of adding `currentColor` to `DrawableParent`, add `currentColor` to `DrawableStyle`. In this way, this property would be available at each drawable through `parent.style`.

How does the first approach sound to you? I am not sure if the second one would make sense as `currentColor` is an inherited property of the parent rather than the style of the parent.

lib/src/svg/theme.dart
bselwe: Yeah, that makes sense. Will remove `colorFilter` from `SvgTheme`.

lib/src/svg/theme.dart
bselwe: I have created a `DefaultSvgTheme` - an inherited widget that provides `SvgTheme` to descendant widgets. 

I am now thinking of a clean way to access the provided theme in `SvgPicture`. The theme needs to be passed to `svg.svgPictureDecoder` (so it can be provided to `SvgParser`) and at the moment all decoders are initialized in `SvgPicture` constructors. To access the theme from context, we could probably move the initialization of decoders to `initState`. Do you think that would be okay?

lib/src/svg/parser_state.dart
dnfield: Perhaps the PR landed after you started looking at this, but there's already support for currentColor to work this way without an extra property - see https://github.com/dnfield/flutter_svg/pull/588

All that's really missing is a way to supply it to the root without it defaulting to black :)

lib/src/svg/theme.dart
dnfield: Yes - all of this plays into the caching concerns I was mentioning. Today we assume that the picture is static and only the width/height will change -but now properties related to decoding can change.

lib/src/svg/parser_state.dart
bselwe: I think I confused how the currentColor property should work. If you provide this value through a widget's property, it should only apply it to the style of the root SVG element, but not to any descendant SVG elements, right?

In the example below, the circle should have a green color rather than the value of `currentColor` provided through widget's property?
```
<svg fill="green"
 width="2em" height="1em" viewBox="0 0 200 100">
    <circle cx="50" cy="50" r="45" fill="currentColor"/>
</svg>
```

bselwe:
Hey @dnfield, I have made the following updates if you could take a look:
- The currentColor property is now supplied only to the root SVG element.
- Picture providers initialization was moved from constructors to `didChangeDependencies` (when `SvgTheme` from context changes) and `didUpdateWidget` (when widget's`PictureProviderMetadata` changes).

lib/src/svg/parser_state.dart
dnfield: The circle would have green, yes.

bselwe: It seems that the `svg` element needs to define a [`color`](https://www.w3.org/TR/SVG11/color.html#ColorProperty) attribute for the circle to be green.

The above code produces this result:
CleanShot 2021-09-29 at 19 37 39@2x

Using `color` for `svg` element (rather than `fill`) produces this result:

CleanShot 2021-09-29 at 19 38 22@2x

lib/src/picture_provider.dart
```diff
@@ -299,7 +305,7 @@ class PictureConfiguration {
 abstract class PictureProvider<T> {
   /// Abstract const constructor. This constructor enables subclasses to provide
   /// const constructors so that they can be used in const expressions.
-  const PictureProvider(this.colorFilter);
+  PictureProvider(this.colorFilter);
```
dnfield, line 308: We should check if this will break any internal clients.

lib/src/svg/xml_parsers.dart
```diff
@@ -178,6 +178,7 @@ DrawablePaint? parseStroke(
   Rect? bounds,
   DrawableDefinitionServer definitions,
   DrawablePaint? parentStroke,
+  Color? currentColor,
```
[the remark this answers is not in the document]
bselwe: It is used [here](https://github.com/dnfield/flutter_svg/pull/602/files/f66024d5832bf83ccf5c20713a490db67e0d36dd#diff-668126c7a12ea8f3c853ca4425f17a33918ff17b0ed2ab78936dc3c512130fceR222) when `parseColor` returns null for the `rawStroke` equal to `currentcolor`/`currentColor`.

lib/src/picture_provider.dart
bselwe: Do you know if there is an easy way to validate that? Regarding the usage of `HttpClient`, from what I see, a new instance is created in each `PictureProvider._loadAsync` call.

lib/src/svg/xml_parsers.dart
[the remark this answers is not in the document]
dnfield: It doesn't look like we're passing that into this function though, are we?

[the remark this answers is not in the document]
bselwe: It is passed in every usage of `parseStyle` in `SvgParserState`: [#125](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR125), [#144](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR144), [#166](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR166), [#193](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR193), [#215](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR215), [#558](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR558), [#931](https://github.com/dnfield/flutter_svg/pull/602/files#diff-4958378e120812df7ee27fb1a0ee4da60f273c5977af2c8aa80483cbb99921ceR931).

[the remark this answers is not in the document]
dnfield: But this function is `parseStroke`, right?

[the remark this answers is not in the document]
dnfield: Ah! Ok, I was confused because of the way the diff collapsed. I see it now.

lib/src/picture_provider.dart
dnfield: I did a quick internal code search and we have one client this would break. It should be easy to fix.

dnfield (approved):

PR #605, opened 2021-10-07
Add support for em units

bselwe:
Adds support for font-relative em units.

- Added `parseDoubleWithUnits` which parses a raw `String` to a `double` taking into account absolute and relative units (px and em).
- Added `fontSize` to `SvgTheme`.
- `SvgPicture` now receives `SvgTheme` as a parameter and uses the provided font size to calculate em units in SVG elements. If no font size is provided through the widget's property, then the value fallbacks to (consecutively) `DefaultSvgTheme.fontSize`, `DefaultTextStyle.style.fontSize` or 14.0.
- The following SVG elements/attributes support em units at the moment:
  - svg (width, height)
  - use (x, y)
  - text (x, y)
  - radialGradient (cx, cy, r, fx, fy)
  - linearGradient (x1, y1, x2, y2)
  - image (x, y, width, height)
  - circle (cx, cy, r)
  - rect (x, y, width, height, rx, ry)
  - ellipse (cx, cy, rx, ry)
  - line (x1, y1, x2, y2)
  - stroke-dasharray
  - stroke-dashoffset
  - stroke-width


bselwe:
@dnfield

bselwe:
@dnfield I have added the following updates:
- A default decoder `PictureInfoDecoder` is created when `PictureProvider` is initialized (as discussed [here](https://github.com/dnfield/flutter_svg/pull/605#discussion_r728284211)).
- `PictureInfoDecoderBuilder` now builds a decoder based on an `SvgTheme` rather than a list of properties (currentColor, fontSize, xHeight, etc.).

Also, it seems that [this test](https://github.com/dnfield/flutter_svg/blob/master/test/widget_svg_test.dart#L544) is failing after some recent changes - when I tried to run it with `useLegacyPaint` set to false it works correctly. Do you have any idea how to fix this?

lib/src/picture_provider.dart
```diff
@@ -460,9 +461,11 @@ abstract class AssetBundlePictureProvider
   PictureInfoDecoder<String> decoder;
 
   @override
-  set currentColor(Color? color) {
-    _currentColor = color;
-    decoder = decoderBuilder(_currentColor);
+  set theme(SvgTheme? theme) {
+    _theme = theme;
```
dnfield, line 465: We should guard this with a check about whether the theme == the other theme or not.

lib/src/picture_provider.dart
```diff
@@ -578,7 +585,8 @@ class NetworkPicture extends PictureProvider<NetworkPicture> {
     }
     return other is NetworkPicture &&
         url == other.url &&
-        colorFilter == other.colorFilter;
+        colorFilter == other.colorFilter &&
+        theme == other.theme;
   }
 
   @override
```
dnfield, line 592: OVerriding hashcode is no longer valid. Argh. If we have a mutable property on here it's going to break the caching logic, which wants to use this as a key.

dnfield:
I've pushed up some fixes that resolve the issue(s) around hashing mutable objects - and added a lint to make sure we don't make that mistake again. Can you fix up the merge conflicts?

Also sorry for the delays - I've been trying to get other changes rolled to internal clients and it's been taking a while :(

lib/src/picture_provider.dart
```diff
@@ -331,18 +331,22 @@ abstract class PictureProvider<T, U> {
...
+  SvgTheme? get theme => _theme;
+  SvgTheme? _theme;
 
-  /// Sets the [_currentColor] to [color].
-  set currentColor(Color? color) {
-    if (_currentColor == color) {
+  /// Sets the [_theme] to [theme].
+  set theme(SvgTheme? theme) {
+    // Fallback to the default theme if null is given.
+    theme = theme ?? const SvgTheme();
```
bselwe, line 342: Would this be correct?

lib/src/picture_provider.dart
```diff
@@ -436,7 +440,7 @@ class PictureKey<T> {
   final ColorFilter? colorFilter;
 
   /// The theme used when this key was created.
-  final SvgTheme theme;
+  final SvgTheme? theme;
```
bselwe, line 443: As the theme is nullable, would this be okay to make this property null in PictureKey as well?

bselwe:
Thanks @dnfield! I have updated this with your latest changes.

lib/src/picture_provider.dart
dnfield: I think we should just make the setter non-nullable then, right?

dnfield: (as in, make the whole property non nullable).

Also, please update the doc string - we should explain what a theme is and how it's used here.

bselwe: Good point, updated.

lib/src/picture_provider.dart
bselwe: Reversed to be non-nullable.

dnfield (approved):

dnfield:
Thanks!

0xba1:
Any chance support for "rem" units, currently parsing "1rem" leads to parsing error: FormatException("Invalid double", source: "1r") @dnfield @bselwe

dnfield:
I think we should probably just treat rem as the same as em, but I'm really not sure. I'd be open to a patch that does that though.

0xba1:
Check out my  PR of doing just that @dnfield https://github.com/dnfield/flutter_svg/pull/646

PR #608, opened 2021-10-13
Fix layer reuse, add regression test

dnfield:
Previous patch was reusing layers. This is invalid, since it leads to the layer being removed when it gets added to a different part of the tree. I thought I had fixed this, but I guess not in whatever I pushed up. Added a regression test.

/cc @goderbauer for any additional thoughts.


dnfield:
Going to land this to try to kick off testing internally. @goderbauer LMK if there is any feedback on this and happy to makemore changes.

PR #611, opened 2021-10-19
Requires Flutter >=2.4.0 for LayerHandle usage

daohoangson:
The Flutter version constraint is out of date. This PR is bumping it to at least 2.4.0.

The commit that landed `LayerHandle` in flutter/flutter  is https://github.com/flutter/flutter/commit/025397ae3f925b316991912977367d3263cc4432.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -21,7 +21,7 @@ dev_dependencies:
 
 environment:
   sdk: '>=2.12.0 <3.0.0'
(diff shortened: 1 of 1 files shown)


daohoangson:
I noticed that you have released v1.0, congratulations 🎉

Looks like the `LayoutHandle` usage is still there, does that mean the performance improvement is significant? Maybe we should merge this PR in order to bump the minimum Flutter version?

dnfield (approved):

PR #618, opened 2021-11-19
Add support for ex units

bselwe:
Adds support for font-relative ex units.

- Modified `parseDoubleWithUnits` to parse ex units along with px and em.
- Added `xHeight` to `SvgTheme` (defaults to the font size divided by 2).
- `SvgPicture` uses the provided x-height to calculate ex units in SVG elements. If no x-height is provided through the widget's property, then the value fallbacks to (consecutively) `DefaultSvgTheme.xHeight` or the font size divided by 2.
- The following SVG elements/attributes support ex units at the moment:
  - svg (width, height)
  - use (x, y)
  - text (x, y)
  - radialGradient (cx, cy, r, fx, fy)
  - linearGradient (x1, y1, x2, y2)
  - image (x, y, width, height)
  - circle (cx, cy, r)
  - rect (x, y, width, height, rx, ry)
  - ellipse (cx, cy, rx, ry)
  - line (x1, y1, x2, y2)
  - stroke-dasharray
  - stroke-dashoffset
  - stroke-width


bselwe:
@dnfield

lib/src/svg/theme.dart
```diff
@@ -22,16 +23,23 @@ class SvgTheme {
...
+  /// The x-height (corpus size) of the font used when calculating ex units of SVG elements.
+  /// Defaults to [fontSize] / 2 if not provided.
+  /// See: https://www.w3.org/TR/SVG11/coords.html#Units, https://en.wikipedia.org/wiki/X-height
+  final double xHeight;
+
   @override
   bool operator ==(dynamic other) {
-    if (other.runtimeType != runtimeType) {
-      return false;
+    if (identical(this, other)) {
```
dnfield, line 33: What's the reason for this change?

This class shouldn't really be expensive enough to compare yet to benefit from this.

test/picture_provider_test.dart
```diff
@@ -108,102 +116,17 @@ void main() {
...
+          xHeight: 6.5,
         );
 
-        expect(exactAssetPicture.decoder, isNotNull);
-        expect(exactAssetPicture.decoder, isNot(equals(decoder)));
-        expect(currentTheme?.currentColor, equals(color));
-      });
-    });
-
-    group(
```
dnfield, line 123: Why are we deleting these?

dnfield (commented):
Overall seems fine, a couple small questions

test/picture_provider_test.dart
bselwe: It's because the decoder is now rebuilt only when PictureProvider's theme changes (before PictureProvider had multiple properties like `fontSize` and `currentColor`, now it's only a `theme` that wraps all those properties). Because of that we only need to check if the decoder rebuilds when the theme changes if that makes sense.

lib/src/svg/theme.dart
bselwe: This method should not be expensive as it would only check whether two references are the same, right? The reason why I wanted to avoid `runtimeType` is that I found `runtimeType.toString` may have different values under minification and saw some recommendations to avoid it ([see here](https://github.com/dart-lang/sdk/issues/35052#issuecomment-621644698)). Let me know if you prefer to keep `runtimeType` here.

dnfield: Yes, runtimeType toString should be avoided but this does not use that.  Please change it back

test/picture_provider_test.dart
dnfield: Ahh ok

lib/src/svg/theme.dart
bselwe: Sure, reverted.

bselwe:
@dnfield I think all your comments should be resolved now. Do you think we can merge this?

dnfield (approved):

dnfield:
Thanks!

PR #619, opened 2021-11-22
Svg as xlink href inside SVG file.

ashish12s3:
@dnfield this PR to support of SVG url inside image tag. Previously it support only for images like jpg png etc. for example:

`<image xlink:href="https://upload.wikimedia.org/wikipedia/commons/b/b4/Chess_ndd45.svg"/>`

### DIFF
--- a/example/assets/simple/svg_xlink_svg.svg
+++ b/example/assets/simple/svg_xlink_svg.svg
@@ -1,0 +1,15 @@
+<!-- Learn about this code on MDN: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/clipPath -->
+<svg viewBox="0 0 120 120" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs>
+        <clipPath id="myClip">
+            <circle cx="30" cy="30" r="20"/>
+            <circle cx="70" cy="70" r="20"/>
+        </clipPath>
+    </defs>
+
+    <g>
+        <rect x="0" y="0" width="120" height="120" fill="#f0f"/>
+        <image xlink:href="https://upload.wikimedia.org/wikipedia/commons/b/b4/Chess_ndd45.svg"/>
+    </g>
+</svg>
+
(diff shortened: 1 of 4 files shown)


dnfield:
This needs tests and documentation. It should also specify a bit more about what you're actually trying to fix.

ashish12s3:
@dnfield  this PR to support of SVG url inside image tag. Previously it support only for images like jpg png etc.

lib/src/svg/parser_state.dart
```diff
@@ -580,28 +580,39 @@ class _Elements {
...
-    if (!parserState._inDefs || !isIri) {
-      parserState.currentGroup!.children!.add(drawable);
+
+    if(href.startsWith('http')&& href.endsWith('.svg')){
+      final DrawableRoot drawable = await resolveSvg(
+          href,
+          parserState.theme,
+          parserState._key,
+          parserState._warningsAsErrors);
+      parserState.currentGroup!.children!.addAll(drawable.children);
```
dnfield, line 590: I don't think this will work correctly for a bunch of possible values in viewBox. I'm concerned about adding support for this feature without having a bunch of cases that test and exercise different options here.

ashish12s3: @dnfield  logically it should work, if SVG parser create drawable for children successfully. can you suggest some use case for those I'll check.

ashish12s3: @dnfield can you please help me here with other cases.

example/assets/simple/svg_xlink_svg.svg
```diff
@@ -0,0 +1,15 @@
...
+    <defs>
+        <clipPath id="myClip">
+            <circle cx="30" cy="30" r="20"/>
+            <circle cx="70" cy="70" r="20"/>
+        </clipPath>
+    </defs>
+
+    <g>
+        <rect x="0" y="0" width="120" height="120" fill="#f0f"/>
+        <image xlink:href="https://upload.wikimedia.org/wikipedia/commons/b/b4/Chess_ndd45.svg"/>
```
dnfield, line 12: If you add a `width` parameteron this element, it does not render correctly.

The defs is also unnecessary here, as is the comment on the first line.

dnfield:
This never actually added a programmatic test and is now out of date with the current repo. Please consider opening an issue to more fully describe what you want to achieve.

PR #621, opened 2021-11-28
Paragraphs: Use longestLine instead of minIntrinsicWidth

### LINKED ISSUE #248: Text Rendering issue (text not positioned correctly)
Rendering this SVG 
SVG results in:
Screen Shot 2019-11-12 at 6 08 06 PM

```dart
SvgPicture.network(
            'https://img.shields.io/badge/platform-flutter%7Cflutter%20web%7Cdart%20vm-ff69b4.svg?style=flat-square')
```

jnschulze:
When calculating the position of a laid out paragraph according to the provided `DrawableTextAnchorPosition`, the paragraph's `minIntrinsicWidth` is used.
For text containing white spaces, however, this yields to wrong results.

### Example 1
(SVG taken from the linked issue)

#### Before
image

#### After
image

#### Before (without space)
image

#### Before (with space)
image

#### After (without space)
image

#### After (with space)
image

Fixes #248


dnfield:
Needs tests to be updated but looks good

jnschulze:
I've just added tests for the `text-anchor` attribute besides updating some golden files but it looks like I'm not able to produce valid files :/

dnfield:
I'll take a look at the test files sometime soon.

I think this is the right thing to do, but rolling this internally will be a bit of a pain so it may take me a little while to land.

dnfield:
(it is likely to break Google internal golden file testing and will require some manual triage work)

dnfield:
Sorry for losing track of this. I'm going to run Google's internal suite of tests against this change to see what the damage is. I haven't forgotten :)

dnfield:
Hm. Somehow this has not affected any golden tests in Google - I think perhaps because using text in SVGs is relatively rare. At any rate that's good news, makes this change a log easier to land. I'm going to see if I can rebase this against master and if tests will pass then.

dnfield:
Thanks for the contribution! I've made some minor changes to your test case to get it to render the way I think you want (or at least to make sure it's more visible if it changes in a bad way). Otherwise this LGTM

PR #646, opened 2022-01-13
Added support for "rem" units

0xba1:


0xba1:
> Please add a test. It's fine if the test just asserts that parseDouble can handle `rem` strings.
> 
> Nitpick: please avoid breaking upt he `else if` with a comment - just combine the comment up over the if, or combine the conditional into a single `if (foo || bar)`.

Thanks for the feedback. New pull request: https://github.com/dnfield/flutter_svg/pull/647

PR #655, opened 2022-01-21
Render performance optimisation suggestion

cristian1980:
Hello,

The condition in set picture(PictureInfo? val) always returns false:
 if (val == picture) {
      return;
 }

This causes markNeedsPaint() to be set more often than it should, happens to me if I zoom the svg in an InteractiveViewer widget. 
Maybe you could implement a == operator in PictureInfo .

Regards,
Cristian

### DIFF
--- a/lib/src/render_picture.dart
+++ b/lib/src/render_picture.dart
@@ -122,7 +122,10 @@ class RenderPicture extends RenderBox {
   PictureInfo? get picture => _picture;
   PictureInfo? _picture;
   set picture(PictureInfo? val) {
-    if (val == picture) {
+    bool b1 = val?.picture == _picture?.picture;
+    bool b2 = val?.size == _picture?.size;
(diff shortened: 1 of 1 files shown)


dnfield (changes requested):
This should probably just be an operator == override on PictureInfo.

I removed those when that object became mutable, but in this case it should be ok to override == and not hash code, or to override them even though it's a mutable object since we have other ways to check if the object is dispsoed.

dnfield:
So I now disagree with myself - overriding operator equals and hash code on pictureInfo gets really messy, and these really are the only rendering properties we care about. But this is formatted incorrectly and should have some commentary explaining why we're doing this.

dnfield:
Closing in favor of https://github.com/dnfield/flutter_svg/pull/656

dnfield:
Thanks for the suggestion!

PR #658, opened 2022-01-25
Update golden images for SkParagraph text rendering

jason-simmons:


dnfield (approved):
Lgtm

The only problem is this will fail on my ci until it lands in beta and/or stable. I can skip it after this commit until that happens

Issue #669, opened 2022-02-24
strokeWidth seems not working with linearGradient


delwinwal:
Hi, i'm trying to render this svg file:
![circle](https://user-images.githubusercontent.com/54469675/155556309-62e639ef-8239-48dc-bf55-6a943657705f.svg)
```xml
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
    <linearGradient id="gradient" gradientUnits="userSpaceOnUse" x1="0" x2="100" y1="50" y2="50">
        <stop offset=".0" stop-color="#0000ff" />
        <stop offset=".3" stop-color="#00ccff" />
        <stop offset=".6" stop-color="#00ffd5" />
        <stop offset=".9" stop-color="#00ff00" />
    </linearGradient>
    <g fill="none" stroke-miterlimit="10">
        <circle cx="50" cy="50" r="40" stroke="url(#gradient)" stroke-linecap="round" stroke-width="10" />
    </g>
</svg>
```
But when I add it in flutter with `SvgPicture.assets`:
```dart
SvgPicture.asset('assets/icons/circle.svg')
```
I got this: 
![image](https://user-images.githubusercontent.com/54469675/155554665-210a3f2a-179b-4a76-b331-60e14118dbf1.png)
while it should look like this:
![image](https://user-images.githubusercontent.com/54469675/155554818-aa62d5d7-9629-4694-8a55-65cd48e38898.png)

It dosen't takes the `stroke-width` parameter into account.
If I `stroke="url(#gradient)"` by a simple hex color, it works correctly.

I've been lookin to this the following posts, but they are from years ago, and none of them seems to have a proper solution/answer:
https://stackoverflow.com/questions/63301612/flutter-svg-files-stroke-becomes-more-thin
https://github.com/dnfield/flutter_svg/issues/551
https://github.com/dnfield/flutter_svg/issues/176

dnfield:
I am able to reproduce this. Taking a look.

dnfield:
I see the problem - if there's a reference to a stroke I'm ignoring any attributes that should be added.

dnfield:
Fix for this incoming, just testing the fix with a larger set of images to make sure it doesn't break things :)

PR #671, opened 2022-03-01
Fix stroke attribute handling when gradients or groups are involved

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/669


dnfield:
Test failures are, unfortunately, expected due to SkParagraph changes that are not yet present in dev/stable. I also need to change CI so that I'm not testing on dev anymore because it won't be updated.

PR #674, opened 2022-03-05
Make sure the SvgPicturestate holds on to a handle

dnfield:
Fixes an issue where a didUpdateWidget can result in disposing the picture too early.


dnfield:
Stable failure expected.

PR #678, opened 2022-03-10
Colors support inherit and opacity

qiongshusheng:
fill attribute not support inherit and opacity， when the fill="Inherit" parseColor function throw exception， and fill="rgba(red, green, blue, opacity)" the opacity will be lost。

### DIFF
--- a/test/xml_svg_test.dart
+++ b/test/xml_svg_test.dart
@@ -1,5 +1,6 @@
 import 'dart:ui';
 
+import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:flutter_svg/src/svg/parser_state.dart';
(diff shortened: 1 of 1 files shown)


lib/src/svg/parser_state.dart
```diff
@@ -1439,6 +1439,7 @@ class SvgParserState {
         defaultFillColor;
 
     if (explicitOpacity && color != null) {
+      opacity *= color.opacity;
```
dnfield, line 1442: This doesn't seem right to me. Can you explain? Is there a test case SVG that is rendering incorrectly without this change but correctly with it?

dnfield:
I'm going to close this - it's not clear to me what this change is supposed to do or that it's safe to make.

PR #689, opened 2022-04-01
Extends SvgTheme to have locale and decideFontFamily.

espresso3389:
For #688, I extended `SvgTheme` to have `locale` and also `decideFontFamily` callback that decide actual font from list of font families listed on `text` element.

This is just a draft PR but I want to discuss how to deal with such cases.

### DIFF
--- a/lib/src/svg/parser_state.dart
+++ b/lib/src/svg/parser_state.dart
@@ -1786,9 +1786,16 @@ class SvgParserState {
     throw StateError('Could not parse "$colorString" as a color.');
   }
 
-  /// Process CSS style font family list to decide a font family.
+  /// Parse CSS style font family list to decide a font family.
   String? parseFontFamily(String? fontFamily) {
-    return theme.decideFontFamily?.call(_parseStringList(fontFamily));
+    final fontFamilies = _parseStringList(fontFamily);
+    if (theme.decideFontFamily != null) {
+      return theme.decideFontFamily!(fontFamilies);
+    }
+    if (fontFamilies == null || fontFamilies.isEmpty) {
+      return null;
+    }
+    return fontFamilies[0];
   }
 
   static List<String>? _parseStringList(String? list) {
@@ -1818,7 +1825,9 @@ class SvgParserState {
       }
     }
     _addIfNotEmptyString(out, sb.toString());
-    return out;
+    return out.isNotEmpty
+        ? out
+        : null; // if the list is empty, return null instead.
   }
 
   static void _addIfNotEmptyString(List<String> list, String s) {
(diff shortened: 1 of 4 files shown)


lib/src/svg/parser_state.dart
```diff
@@ -1783,6 +1785,57 @@ class SvgParserState {
 
     throw StateError('Could not parse "$colorString" as a color.');
   }
+
+  /// Parse CSS style font family list to decide a font family.
+  String? parseFontFamily(String? fontFamily) {
+    final fontFamilies = _parseStringList(fontFamily);
+    if (theme.resolveFontFamily != null) {
+      return theme.resolveFontFamily!(fontFamilies);
```
dnfield, line 1793: What's the use case here?

Why would we avoid using the `fontFamilyFallack` on `TextStyle` to append any additional fonts?

lib/src/svg/parsers.dart
```diff
@@ -191,8 +191,9 @@ Paragraph createParagraph(
   String text,
   DrawableStyle style,
   DrawablePaint? foregroundOverride,
+  Locale? locale,
 ) {
-  final ParagraphBuilder builder = ParagraphBuilder(ParagraphStyle())
+  final ParagraphBuilder builder = ParagraphBuilder(ParagraphStyle(locale: locale))
```
dnfield, line 196: Perhaps
```suggestion
  final ParagraphBuilder builder = ParagraphBuilder(ParagraphStyle(locale: locale ?? PlatformDispatcher.instance.locale))
```

lib/src/svg/parser_state.dart
espresso3389: Yes, appending them to `fontFamilyFallack` may help but `font-family` on `text` may have many variations and they cannot be recognized by the system... How should I...

For my own use case, a SVG file that has

```html
<text fill="#fff" font-family="YuGo-Medium-83pv-RKSJ-H, YuGothic" font-size="4px" transform="translate(6.63 22.25)">
```

There are two font names here; `YuGo-Medium-83pv-RKSJ-H` and `YuGothic`.

And, it seems that iOS/macOS can resolve the font name `YuGothic` but Windows can't. There should be `Yu Gothic`.

Therefore, in my app, I just made a hand-crafted mapping table and translate such variations.

dnfield: Why not update your SVG to include `Yu Gothic` in the `font-family`? :)

espresso3389: Yes, it's sometimes the correct and best way to update SVG to include both "YuGothic", "Yu Gothic", and also "游ゴシック". But it's not acceptable if we have many files...

espresso3389: Anyway, I prefer some dynamic way to resolve font names based on installed fonts or platform name, or... So it's better to equip a away to resolve such kind of mechanism.

dnfield:
This PR is out of date with the main branch, closing for now.

PR #701, opened 2022-04-30
Remove redundant canvas.restore()

FaFre:
The canvas is getting restored in some scenarios one time too much, which will lead to an exception and invalid canvas.

The value of `viewBoxOffset` does not have any effect on saving the canvas initially. The `save()`  is controlled by `transform != null` and eventually restored in L954.

### DIFF
--- a/test/vector_drawable_test.dart
+++ b/test/vector_drawable_test.dart
@@ -108,9 +108,11 @@ void main() {
     final PictureRecorder recorder = PictureRecorder();
     final Canvas canvas = Canvas(recorder);
 
+    canvas.save();
+
     final DrawableRoot svgRoot = await svg.fromSvgString(
       svgWithTransform,
-      'RestoieCanvasWithTransform',
+      'RestoreCanvasWithTransform',
     );
 
     svgRoot.scaleCanvasToViewBox(canvas, const Size.square(200));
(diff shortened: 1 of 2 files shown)


dnfield (approved):
Ahh, good. Can you add a test for this? It should be testable if you call draw and check the save count on the canvas at the end.

FaFre:
> Ahh, good. Can you add a test for this? It should be testable if you call draw and check the save count on the canvas at the end.

done!

test/vector_drawable_test.dart
```diff
@@ -88,4 +88,38 @@ void main() {
...
+      svgWithTransform,
+      'RestoieCanvasWithTransform',
+    );
+
+    svgRoot.scaleCanvasToViewBox(canvas, const Size.square(200));
+    svgRoot.clipCanvasToViewBox(canvas);
+
+    svgRoot.draw(canvas, svgRoot.viewport.viewBoxRect);
+
+    expect(canvas.getSaveCount(), equals(1));
```
dnfield, line 121: Unfortunately I was wrong - getSaveCount can never go below 1, so this test passes without your change.

I think the easiest way to do this would be to add a save before you pass the canvas in and check that it still equals 2 here and before passing it in.

dnfield: IOW:

On line 110, add `canvas.save(); expect(canvas.getSaveCount(), 2);` and here expect that it's 2 again.

FaFre: @dnfield I fixed the test

FaFre: I also reduced the SVG code, as we actually don't need to draw anything to validate the behavior. Makes things clearer.

dnfield (approved):

Issue #708, opened 2022-05-18
Upgrade dependancy to xml 6.0.1


guchengxi1994:

mariopepe:
+1

dnfield:
Done on master

PR #713, opened 2022-05-19
Update project to support flutter 3.0

erdemyerebasmaz:
This PR addresses #708.

Most plugins are migrating to Flutter 3.0 and there has been version solving incompatibilities due to flutter_svg using xml ^5.0.0.

#### Changes:
- Updated plugins to latest resolvable
  - There were no breaking changes according to their changelogs published in pub.dev
- Dart requirement set to 2.17
- Removed deprecated splash screen. ([See here](https://docs.flutter.dev/development/ui/advanced/splash-screen#migrating-from-manifest--activity-defined-custom-splash-screens))
- Removed deprecated **new** keyword in comment reference. ([See here](https://dart.dev/tools/diagnostic-messages#deprecated_new_in_comment_reference))

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -6,21 +6,21 @@ version: 1.1.0
 dependencies:
   flutter:
     sdk: flutter
-  meta: ^1.3.0
+  meta: ^1.7.0
   path_drawing: ^1.0.0
-  vector_math: ^2.1.0
+  vector_math: ^2.1.2
 
-  xml: ^5.0.0
+  xml: ^6.0.1
 
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  path: ^1.8.0
-  test: ^1.16.0
+  path: ^1.8.1
+  test: ^1.21.1
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.4.0-0.0.pre"
+  sdk: '>=2.17.0 <3.0.0'
+  flutter: ">=3.0.0"
 
 ## For developing features that require new functionality in engine:
 # dependency_overrides:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -19,7 +19,7 @@ dev_dependencies:
   test: ^1.21.1
 
 environment:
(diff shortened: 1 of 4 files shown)


dnfield (approved):

dnfield:
Thanks!

pubspec.yaml
```diff
@@ -6,21 +6,21 @@ version: 1.1.0
...
     sdk: flutter
-  path: ^1.8.0
-  test: ^1.16.0
+  path: ^1.8.1
+  test: ^1.21.1
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.4.0-0.0.pre"
+  sdk: '>=2.17.0 <3.0.0'
```
dnfield, line 22:
```suggestion
  sdk: '>=2.17.0-0 <3.0.0'
```

pubspec.yaml
```diff
@@ -6,21 +6,21 @@ version: 1.1.0
...
-  path: ^1.8.0
-  test: ^1.16.0
+  path: ^1.8.1
+  test: ^1.21.1
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
-  flutter: ">=2.4.0-0.0.pre"
+  sdk: '>=2.17.0-0 <3.0.0'
+  flutter: ">=3.0.0"
```
dnfield, line 23: This seems to make CI unhappy. I think we can just leave it at the 2.4 with no upper bound. WDYT?

erdemyerebasmaz: Splash screen change was deprecated after 2.5 so I think it should be the lower bound.

erdemyerebasmaz: Upon review, the CI error appears to be in beta channel.
I propose we use `2.11.0-0.1.pre` as lower bound as it's the version that switches over to Dart 2.17.

dnfield:
Thanks!

mariopepe:
Is the version 1.1.0 being verified from Google? At the moment on pub.dev the most recent one is still 1.0.3.

Issue #720, opened 2022-06-16
When using flutter 3 then some svg not rendered at all


Lufan:
I have project with several svg that loos like
```svg
<svg
  xmlns="http://www.w3.org/2000/svg"
  width="24"
  height="24"
  viewBox="0 0 24 24"
  fill="none"
  stroke="currentColor"
  stroke-width="2"
  stroke-linecap="round"
  stroke-linejoin="round"
>
  <path d="M3 9l9-7 9 7v11a2 2 0 01-2 2H5a2 2 0 01-2-2z" />
  <polyline points="9 22 9 12 15 12 15 22" />
</svg>
```

when I change in the line '_stroke="currentColor"_' value '_currentColor_' to some concrete color like '_black_' - then svg rendered fine.

This svg reported as valid when i check it with https://validator.w3.org/check.

With previous versions of the flutter it works well, only stop working when i update flutter to version 3.0.1 or 3.0.2 and apply flutter clean to the project

to reproduce this i used simplified version of example:
where dart.svg - i get from this repository and it rendered fine, but home.svg not rendered 
```dart
import 'package:flutter/material.dart';
import 'package:flutter_svg/avd.dart';
import 'package:flutter_svg/flutter_svg.dart';

void main() {
  runApp(_MyApp());
}

class _MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const _MyHomePage(title: 'Flutter SVG Demo'),
    );
  }
}

class _MyHomePage extends StatefulWidget {
  const _MyHomePage({Key? key, required this.title}) : super(key: key);
  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

class _MyHomePageState extends State<_MyHomePage> {
  final List<Widget> _painters = <Widget>[];
  late double _dimension;

  @override
  void initState() {
    super.initState();
    _dimension = 203.0;
    _painters.add(
      Directionality(
        textDirection: TextDirection.ltr,
        child: SvgPicture.asset(
          'assets/images/home.svg',
          color: Colors.black
        ),
      ),
    );
    _painters.add(
      Directionality(
        textDirection: TextDirection.ltr,
        child: SvgPicture.asset(
          'assets/images/dart.svg',
        ),
      ),
    );
  }

  @override
  Widget build(BuildContext context) {
[...]

PR #722, opened 2022-06-18
Restored 'color' attribute validation to render svg image correctly with stroke="currentColor" on <svg>

Lufan:
Changes to resolve issue #720


dnfield:
CI is messed up right now for other reasons, and this needs a test. Otherwise this looks fine to me. I'm happy to take this over - thanks for the suggestion, will land something like this today.

Issue #743, opened 2022-08-04
[Parser] Throw exception when parse attribute '1pt'


Saladin-Khanhnt:
An exception was throw when try to draw image
`<svg
	xmlns="http://www.w3.org/2000/svg" id="flag-icons-id" viewBox="0 0 512 512">
	<g fill-rule="evenodd" stroke-width="1pt">
		<path fill="#e70011" d="M0 0h512v265.5H0z"/>
		<path fill="#fff" d="M0 256h512v256H0z"/>
	</g>
</svg>`

Version: flutter_svg: ^1.1.2
Exception stack trace:
======== Exception caught by SVG ===================================================================
The following FormatException was thrown resolving a single-frame picture stream:
Invalid double
1pt

When the exception was thrown, this was the stack: 
#0      double.parse (dart:core-patch/double_patch.dart:111:28)
#1      parseDouble (package:flutter_svg/src/utilities/numbers.dart:23:17)
#2      SvgParserState.parseDoubleWithUnits (package:flutter_svg/src/svg/parser_state.dart:1067:27)
#3      SvgParserState.parseStroke (package:flutter_svg/src/svg/parser_state.dart:1372:20)
#4      SvgParserState.parseStyle (package:flutter_svg/src/svg/parser_state.dart:1608:15)
#5      _Elements.g (package:flutter_svg/src/svg/parser_state.dart:154:19)
#6      SvgParserState.parse (package:flutter_svg/src/svg/parser_state.dart:887:26)

leecasey:
Hi sorry, has this issue been resolved now? will there be an update soon?

leecasey:
All working again with 1.1.3 - thank you! Amazing support.

PR #744, opened 2022-08-04
Fixed issue #743: [Parser] Throw exception when parse attribute '1pt'

khanhnt748:


lib/src/utilities/numbers.dart
```diff
@@ -15,6 +15,7 @@ double? parseDouble(String? rawDouble, {bool tryParse = false}) {
       .replaceFirst('em', '')
       .replaceFirst('ex', '')
       .replaceFirst('px', '')
+      .replaceFirst('pt', '')
```
dnfield, line 18: So I think I've been overthinking this.

I think we can just use CSS's rule here and just multipy `pt` values by 1.33333 to get pixels. This isn't quite the right fix though.

dnfield: Can you take a look at https://github.com/dnfield/vector_graphics/pull/114/files?

We'd want a similar fix here.

dnfield: Are you interested in doing this? If not I'll push up a fix.

dnfield:
I'm going to go ahead and push up a fix for this in a new PR. Thanks!

PR #748, opened 2022-08-16
Don't require defs to be put ahead

ikbendewilliam:
Don't require the defs to be put first, this will sort before handling the svg. Fixing at least some cases of #102


dnfield:
Obsoleted by

Issue #749, opened 2022-08-17
Unable to display SVGs with <g><image(base64)>


AlexV525:
The preview is entirely empty.

Source SVG: https://qbc6i-daaaa-aaaah-qcywq-cai.raw.ic0.app/?tokenid=lrmer-zqkor-uwiaa-aaaaa-b4awf-uaqca-aaavk-q

dnfield:
Pretty sure the bug here is that flutter_svg is expecting there to be `width`/`height` attributes, which aren't necessarily required by the spec.

PR #750, opened 2022-08-17
Default image width/height correclty

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/749 (@alexv525).

@jonahwilliams fyi


dnfield:
Golden diff is acceptable and will need to update goldens for it.

PR #756, opened 2022-08-27
Defs and refs

### LINKED ISSUE #102: Out of order defs/references
Right now, an error is reported if you try to use a file that does something like the following:

```svg
<svg>
  <path fill="url(#MyGradient)" d="..." />
  <linearGradient id="MyGradient" ...>
    ...
  </linearGradient>
</svg>
```

This is done to make sure we don't have to maintain a full DOM or traverse the XML more than once.

However, it should be possible to keep an in memory structure of defs (probably a `Map` of some sort) we expect to see but haven't seen yet, and go back and back-fill them (and then remove them from the structure after we found the def they're looking for) with only a little bit of memory overhead.

See also #68  and #97

ikbendewilliam:
Fixes #102 
Defs and refs can now be used anywhere, the parsing is now partially async and will await for the def/ref to be defined or throw a warning (like previously) when not found and the full file is processed.
This is a very different approach than #748 which only solves defs and has a lower impact
This is also not (yet) compatible with #754 
I haven't done a benchmark yet, so I don't know the full impact yet, but I've tried to minimise it. Let me know if this is something you expect 😄.

### DIFF
--- a/example/assets/flutter_logo.svg
+++ b/example/assets/flutter_logo.svg
@@ -1,4 +1,12 @@
 <svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 166 202">
+    <path fill="#42A5F5" fill-opacity=".8" d="M37.7 128.9 9.8 101 100.4 10.4 156.2 10.4"/>
+    <path fill="#42A5F5" fill-opacity=".8" d="M156.2 94 100.4 94 79.5 114.9 107.4 142.8"/>
+    <path fill="#0D47A1" d="M79.5 170.7 100.4 191.6 156.2 191.6 156.2 191.6 107.4 142.8"/>
+    <g transform="matrix(0.7071, -0.7071, 0.7071, 0.7071, -77.667, 98.057)">
+        <rect width="39.4" height="39.4" x="59.8" y="123.1" fill="#42A5F5" />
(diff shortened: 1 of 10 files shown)


dnfield:
I haven't read this whole PR, but a couple comments:

- This is making a lot of unrelated changes. In particular, dropping "unnecessary" asserts should not be done yet - it's not really a good idea to do until sound null-safety is the only option in the Dart VM, which is on the roadmap but a little ways away. The problem is that programs running in mixed mode still may benefit from these asserts.
- I'm concerned about making changes to this library that will make parsing more expensive. Parsing takes a _very_ long time right now and it's a problem for performance sensitive customers as it is. 
- I have a plan to potentially get rid of all of this logic and just make this package use vector_graphics_compiler (but probably without all optimizations) under the hood.

ikbendewilliam:
@dnfield Thank you for your feedback. 
1. Right, you are correct! That makes sense then why it was still there
2. I knew this would be your response. Nothing we can do about it :-) I am planning to do a benchmark, but it's a long shot to get this merged seeing as there are quite a few drawbacks. (speed and complexity being the main ones)
3. I agree that that would be optimal. People that need the optimizations should then use the compiler and the binary format.

dnfield:
Obsolete - this feature now works after some major refactoring.

PR #775, opened 2022-09-13
mod: updated contribting.md file

seths10:


dnfield:
Thanks, but I'm not sure how much this really helps :)

PR #782, opened 2022-10-03
Start using VG

### LINKED ISSUE #102: Out of order defs/references

### LINKED ISSUE #289: Support the <pattern> element
`<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" height="273" width="298">
  <defs>
    <image id="1a54d514-63f6-47f8-a0f4-ed80dsdddsf04a" `

Are this type of images supported ?
And if how to  use them?

### LINKED ISSUE #736: Failed to find definition for url(#paint0_linear_883_8164)
I failed to render the svg file completely with flutter svg, because some images are missing, like the image below
**Image**
Screen Shot 2022-07-15 at 16 05 31

error message
`═══════ Exception caught by SVG ═══════════════════════════════════════════════
The following assertion was thrown while parsing AssetBundlePictureKey(bundle: PlatformAssetBundle#46450(), name: "assets/images/splash.svg", colorFilter: null, theme: SvgTheme(currentColor: Color(0xff000000), fontSize: 48.0, xHeight: 24.0)) in _getDefinitionPaint:
Failed to find definition for url(#paint0_linear_883_8164)

This library only supports <defs> and xlink:href references that are defined ahead of their references.

This error can be caused when the desired definition is defined after the element referring to it (e.g. at the end of the file), or defined in another file.

This error is treated as non-fatal, but your SVG file will likely not render as intended
════════════════════════════════════════════════════════════════════════════════

════════ Exception caught by SVG ═══════════════════════════════════════════════
Failed to find definition for url(#pattern0)
════════════════════════════════════════════════════════════════════════════════

════════ Exception caught by SVG ═══════════════════════════════════════════════
Failed to find definition for url(#pattern1)
════════════════════════════════════════════════════════════════════════════════`

**svg file**
https://drive.google.com/file/d/1RImrm5gwX9kHWycs2eKOvx6Ehddos1Z9/view?usp=sharing

can you give me some solutions?

dnfield:
Fixes #736
Fixes #289 
Fixes #102


dnfield:
Some things to do:

- [ ] Port some tests that will no longer make sense here to vector_graphics
- [ ] Get support for font-size and current-color specification into the compiler/loader spec (i.e. what SvgTheme is handling now, but not necessarily with the SvgTheme class).
- [ ] Document the new loaders.
- [ ] Update docs/changelog

dnfield:
Color filter change might be slightly problematic. Otherwise, I think this is more or less in good shape. @jonahwilliams fyi

dnfield:
I'll see what an internal test run says on this for number of goldens changed.

dnfield:
Actually, this will probbly have failures on internal usages of the `VectorDrawable` API. So it'll take a little more than a simple run.

jonahwilliams:
What should I be looking at for color filter?

dnfield:
The way the bounds are getting calculated seem different. Might be ok.

jonahwilliams:
Ahh yeah, I'll double check that

jonahwilliams:
Nothing obvious yet, perhaps a difference in how saveLayer and ColorFilterLayer are implemented. I did find https://github.com/dnfield/vector_graphics/pull/152 though

lib/src/loaders.dart
```diff
@@ -0,0 +1,213 @@
...
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/src/utilities/http.dart';
+import 'package:vector_graphics/vector_graphics.dart';
+import 'package:vector_graphics_compiler/vector_graphics_compiler.dart';
+
+/// Compute will mess up tests because of FakeAsync.
+final bool _isTest = Platform.executable.endsWith('flutter_tester') ||
```
jonahwilliams, line 11: We should guard this with a kDebugMode check

jonahwilliams: and probably a kIsWeb check too

lib/src/loaders.dart
```diff
@@ -0,0 +1,213 @@
...
+
+  /// The file containing the SVG data to decode and render.
+  final File file;
+
+  /// The theme to determine currentColor and font sizing attributes.
+  final SvgTheme theme;
+
+  @override
+  Future<ByteData> loadBytes(BuildContext context) async {
+    final Uint8List bytes = file.readAsBytesSync();
```
jonahwilliams, line 100: Since this is already async I would the async file API

jonahwilliams (commented):
Because of the large cost of parsing from source I think you should leave the picture caches and wire them back up for anything that doesn't work from the binary encoded format

dnfield:
> Because of the large cost of parsing from source I think you should leave the picture caches and wire them back up for anything that doesn't work from the binary encoded format

I think I'd rather just cache the bytes from compute.

lib/src/loaders.dart
dnfield: Why kIsWeb?

lib/src/loaders.dart
dnfield: I always avoid it and I'm not entirely sure why, but I believe it's generally more expensive on the implementation side right?

lib/src/loaders.dart
jonahwilliams: Platform.whatever aren't supported

dnfield:
Caching is such a pain though. I'll have to think about this a little.

lib/src/loaders.dart
jonahwilliams: For things like exists, the sync version is much faster. But for reading the bytes you're potentially blocking the UI isolate for multiple milliseconds if disk I/O is getting used elsewhere.

dnfield: Ahhh I meant to put this in the compute isolate

jonahwilliams: Yeah that's much better, we can avoid copying the bytes an extra time too

dnfield:
Latest update: there's usage in google3 of the VectorDrawable classes to add drop shadows in the shape of what's drawn in the SVG. I have a rough prototype of supporting that in this package and/or vector_graphics, need to clean it up.

dnfield:
This is getting pretty close. I'm now down to reviewing golden image changes in Google's monorepo and trying to make sure they're only changes we'd expect with v_g (gradients being a little different, rrects/circles being slightly different, occasional other minor pixel differences).

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+  - Out of order defs now supported.
+  - Patterns supported.
+  - Many optimizations added.
+  - **REGRESSION**: Lost support for `dx` and `dy` on `text`. See
+    https://github.com/dnfield/vector_graphics/issues/44 if this is important to
+    you.
+- Parse SVG data in background isolate by default.
+- Use widgets/rendering strategy from `vector_graphics`.
+- Much less caching.
+- More tests work without special handling - most of parsing is now async,
```
jonahwilliams, line 15: Sync?

dnfield: Yes.

dnfield:
@jonahwilliams only question remaining IMHO is whether to do any caching. I think the only place it could end up hurting is network assets that aren't already getting cached elsewhere.

jonahwilliams:
I'd worry about cases like lazy scrolling constantly re-downloading assets as you scrolled through. I think we should have some sort of cache for the loaders that do compliation because of the cost

larry-cariq:
@dnfield Thanks for updating this package. When will the 2.0.0 version be released on pub.dev?

jonahwilliams:
I will publish a pre-release version of this package later this week.

felipecastrosales:
News about the release of the version? I'm excited to use :D

felipecastrosales:
News about the release of the version? I'm excited :D 
cc @dnfield and @jonahwilliams.

jonahwilliams:
Actually only @dnfield has publishing rights :)

felipecastrosales:
@dnfield, Is there any expected release date for this version with these fixes?

elodieferrais:
@dnfield, any update on a release date for these fixes?

dnfield:
It's published as version 2.x

PR #785, opened 2022-10-10
Add errorBuilder

tak8997:
This is not a complete PR, but how about handling errorBuilder with this concept?


dnfield:
This would have to get done in package:vector_graphics now.

PR #793, opened 2022-10-20
default value instead of null attribute

kirill09:
some svg have null attributes return the default value


dnfield:
This would have to be done in package:vector_graphics_compiler now.

PR #794, opened 2022-10-20
Transform clip path when applied to a group

### LINKED ISSUE #136: <clipPath> with a `transform` won't work.
So, i'm trying to render svg to screen using as usual `SvgPicture.asset()` but it doesn't show anything or even render.

There also no error thrown at the debug console, it just doesn't render

[An example of the svg](https://gist.github.com/ragasubekti/d3e03a76ae7f60bd8bee66430ce2fe7f)

I've also tried to use `usvg` but the result still same

FeodorFitsner:
Fix #136, #343, #566


FeodorFitsner:
Thank you!

PR #798, opened 2022-10-22
Fix: Strokes with zero width should not be painted

### LINKED ISSUE #797: Strokes with zero width should not be painted
Consider the following SVG:

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" class="main-svg" width="100" height="100" viewBox="0 0 100 100">
    <rect style="stroke: rgb(68, 68, 68); stroke-opacity: 1; fill: rgb(255, 255, 255); fill-opacity: 1; stroke-width: 0px;" width="90" height="90" x="5" y="5" />
</svg>
```

According to [`stroke-width` specs](https://www.w3.org/TR/svg-strokes/#StrokeWidth) ***A zero value causes no stroke to be painted***.

However, in Flutter a stroke with zero width [***correspond to a hairline width***](https://api.flutter.dev/flutter/dart-ui/Paint/strokeWidth.html), so the SVG above is displayed like that:

image

Expected result: the line is not visible.

FeodorFitsner:
Fix #797


dnfield (approved):

dnfield:
Thanks!

PR #817, opened 2022-12-11
Avoid anon closure

dnfield:
@jonahwilliams

### DIFF
--- a/lib/src/loaders.dart
+++ b/lib/src/loaders.dart
@@ -35,25 +35,27 @@ abstract class SvgLoader<T> extends BytesLoader {
   Future<T?> prepareMessage(BuildContext? context) =>
       SynchronousFuture<T?>(null);
 
+  Future<ByteData> _load(BuildContext? context) {
+    return prepareMessage(context).then((T? message) {
+      return compute((T? message) {
+        return encodeSvg(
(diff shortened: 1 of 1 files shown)


lib/src/loaders.dart
```diff
@@ -35,25 +35,27 @@ abstract class SvgLoader<T> extends BytesLoader {
...
-            colorMapper: colorMapper,
-            debugName: 'Svg loader',
-            enableClippingOptimizer: false,
-            enableMaskingOptimizer: false,
-            enableOverdrawOptimizer: false,
-          ).buffer.asByteData();
-        }, message, debugLabel: 'Load Bytes');
-      });
-    });
+    return svg.cache.putIfAbsent(cacheKey(context), () => _load(context));
```
jonahwilliams, line 58: Isn't this still an anonymous closure?

dnfield: Hm. Yes. But it doesn't give me the runtime error

jonahwilliams (approved):

lib/src/loaders.dart
dnfield: I'm confused about why but uh also have other things going on haha

PR #818, opened 2022-12-16
Use strict analysis modes instead of deprecated 'strong-mode' options

srawlins:
See https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks for more info.


dnfield:
This might need to get cherry picked into g3. Rolling this package as is right now requires extra patching a lot of golden changes.

PR #833, opened 2023-01-26
README: Remove `AssetLoader()` from the precompiled asset example

pento:
I'm not sure why `AssetLoader()` is included in this sample, but I found that `AssetBytesLoader()` returns the `BytesLoader` that `SvgPicture` expects to receive.


dnfield (approved):
LGTM, thanks!

Issue #852, opened 2023-02-10
Flutter SVG | Compare Asset Testing


felipecastrosales:
Hi! 

Before the version `2.0.0`, I had this test that worked fine and made the assertion I wanted. 

```dart
final findAsset = find.byWidgetPredicate(
  (widget) =>
      widget is SvgPicture &&
      (widget.pictureProvider as ExactAssetPicture)
              .assetName
              .compareTo('assets/flutter.svg') ==
          0,
);

expect(findAsset, findsOneWidget);
```

**However**, after `2.0.0` this testis invalid because there were internal change in the package, e.g:
- pictureProvider: 
`The getter 'pictureProvider' isn't defined for the type 'SvgPicture'.
Try importing the library that defines 'pictureProvider', correcting the name to the name of an existing getter, or defining a getter or field named 'pictureProvider'.`
- ExactAssetPicture: 
`The name 'ExactAssetPicture' isn't a type, so it can't be used in an 'as' expression.
Try changing the name to the name of an existing type, or creating a type with the name 'ExactAssetPicture'.`

I would be like to know which is the best option to follow, with the new version of the package. 

I also made a **Minimal Reproducible Example**: 

- https://github.com/felipecastrosales/svg_mre

felipecastrosales:
I found a solution for this. 

Just do it this way: 

```dart
final findAsset = find.byWidgetPredicate(
  (widget) =>
      widget is SvgPicture &&
      (widget.bytesLoader as SvgAssetLoader) // Here! 
              .assetName
              .compareTo('assets/flutter.svg') ==
          0,
);
```

Is this an interesting approach? I believe so. I'll leave the issue open as I'm open to feedback and if there's something better to be done here. And I also made the commit.

dnfield:
Yes, this is the correct way to fix this issue.

felipecastrosales:
Ok, @dnfield. Thanks for answer. :)

dkbast:
This would be great as part of a migration guide - btw there are still references to the picture provider in the code documentation.

Issue #828, opened 2023-01-19
Migration guide for `color` is deprecated.


eli1stark:
In the version 2 of `flutter_svg` the `color` argument is deprecated, what to use instead? `ColorFilter`, if so, which one?

Three of them visually look the same:
`ColorFilter.mode(color, BlendMode.srcIn)`
`ColorFilter.mode(color, BlendMode.modulate)`
`ColorFilter.mode(color, BlendMode.srcATop)`

dnfield:
Thanks, updated the readme. Use srcIn for the example.

dnfield:
The reason I did this is that Flutter has added more advanced color filtering possibilities since this library was written, e.g. to do matrix based filters.

PR #861, opened 2023-02-13
documentation: Adds migration advice, removes no-op deprecations

dkbast:
documentation: Add link to migration guide and remove no-op deprecations and add deprecation messages with suggestions on how to change the code.

Adds the tipps from #852 to the migration document and adds #828 to the deprecation notice so you don't have to look around for the fix.

### DIFF
--- a/lib/svg.dart
+++ b/lib/svg.dart
@@ -29,29 +29,13 @@ final Svg svg = Svg._();
 class Svg {
   Svg._();
 
-  /// A global override flag for [SvgPicture.cacheColorFilter].
-  ///
-  /// If this is null, the value in [SvgPicture.cacheColorFilter] is used. If it
-  /// is not null, it will override that value.
-  @Deprecated('This no longer does anything.')
-  bool? cacheColorFilterOverride;
-
   /// The cache instance for decoded SVGs.
   final Cache cache = Cache();
 }
 
-// ignore: avoid_classes_with_only_static_members
-/// Deprecated class, will be removed, does not do anything.
-@Deprecated('This feature does not do anything anymore.')
-class PictureProvider {
-  /// Deprecated, use [svg.cache] instead.
-  @Deprecated('Use svg.cache instead.')
-  static Cache get cache => svg.cache;
-}
-
 /// A widget that will parse SVG data for rendering on screen.
 class SvgPicture extends StatelessWidget {
-  /// Instantiates a widget that renders an SVG picture using the `pictureProvider`.
+  /// Instantiates a widget that renders an SVG picture using a [BytesLoader].
   ///
   /// Either the [width] and [height] arguments should be specified, or the
   /// widget should be placed in a context that sets tight layout constraints.
@@ -87,8 +71,6 @@ class SvgPicture extends StatelessWidget {
     this.semanticsLabel,
     this.excludeFromSemantics = false,
     this.theme = const SvgTheme(),
-    @deprecated Clip? clipBehavior,
-    @deprecated bool cacheColorFilter = false,
   }) : super(key: key);
 
   /// Instantiates a widget that renders an SVG picture from an [AssetBundle].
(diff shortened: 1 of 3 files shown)


lib/svg.dart
```diff
@@ -29,29 +29,13 @@ final Svg svg = Svg._();
 class Svg {
   Svg._();
 
-  /// A global override flag for [SvgPicture.cacheColorFilter].
-  ///
-  /// If this is null, the value in [SvgPicture.cacheColorFilter] is used. If it
-  /// is not null, it will override that value.
-  @Deprecated('This no longer does anything.')
-  bool? cacheColorFilterOverride;
```
dnfield, line 37: Please don't remove these. There are usages of this API in some customers and removing them makes it much harder to migrate those customers to the new version.

lib/svg.dart
```diff
@@ -87,8 +71,6 @@ class SvgPicture extends StatelessWidget {
     this.semanticsLabel,
     this.excludeFromSemantics = false,
     this.theme = const SvgTheme(),
-    @deprecated Clip? clipBehavior,
```
dnfield, line 90: Ditto here. I'd like to keep these deprecated properties around for a while to ease migration.

lib/svg.dart
```diff
@@ -182,11 +164,16 @@ class SvgPicture extends StatelessWidget {
...
+
+    /// The [ColorFilter] to apply to the picture. An easy way to add color.
+    /// e.g. `ColorFilter.mode(Colors.red, BlendMode.srcIn)`
     ui.ColorFilter? colorFilter,
-    @deprecated ui.Color? color,
-    @deprecated ui.BlendMode colorBlendMode = ui.BlendMode.srcIn,
-    @deprecated Clip? clipBehavior,
-    @deprecated bool cacheColorFilter = false,
+    @Deprecated('Use colorFilter: ColorFilter.mode(color, blendMode) instead '
+        '- default BlendMode is BlendMode.srcIn )')
```
dnfield, line 172:
```suggestion
        '- default BlendMode is BlendMode.srcIn.')
```

dnfield: (ditto below)

lib/svg.dart
```diff
@@ -182,11 +164,16 @@ class SvgPicture extends StatelessWidget {
     this.semanticsLabel,
     this.excludeFromSemantics = false,
     this.theme = const SvgTheme(),
+
+    /// The [ColorFilter] to apply to the picture. An easy way to add color.
+    /// e.g. `ColorFilter.mode(Colors.red, BlendMode.srcIn)`
```
dnfield, line 169: Here and elsewhere: docs should contain complete sentences thatstart with a capital letter and end with a period.

It would be great to expand this a bit to mention that it is also possible to use the `ColorMapper` property which avoids filtering at runtime and instead replaces colors at parse time.

lib/svg.dart
dkbast: Hmm ... I would actually argue that they actually hinder migration, since by keeping them as noop it's not safe to assume that we can "ignore" deprecation warnings as in this case the deprecation already happened.

dkbast:
Hi @dnfield - thanks for the quick review! I just did a migration on a bigger project and having many no-ops around made it a lot more difficult to migrate. 
It was hard to tell if I can still use a deprecated field or if it just doesn't do anything anymore - If they would still have any effect I would 100% agree, but '@deprecated' implies that this is still working but will not in the future.

From the docs: "The intent of the @Deprecated annotation is to inform authors who are currently using the feature, that they will soon need to stop using that feature in their code, even if the feature is currently still working correctly."

Please let me know if you still want me to reintroduce the dead code :) - no offence meant, the migration just did not go as smooth as you probably hoped to achieve

lib/svg.dart
dkbast: Especially with the doc comment still pointing towards this "doing something" this is quite confusing.

felipecastrosales:
Very good, @dkbast. Congratulations on being proactive.

PR #880, opened 2023-03-12
Add svg finder

SimonWeidemann:
This pull request adds a test helper to find  svgs in widget tests.
The implementation is similar to the default find.image function.


pubspec.yaml
```diff
@@ -7,13 +7,12 @@ version: 2.0.3
 dependencies:
   flutter:
     sdk: flutter
+  flutter_test:
+    sdk: flutter
```
dnfield, line 11: This has to get moved to another package. We can't have the runtime depending on flutter_test.

SimonWeidemann: Sorry for the late response.
I see what you mean. 
How should we go further?
Should I create  a separate  repository or should we integrate the package in this one?
For example like this:
```
flutter_svg
│ 
├──.github
├──.gitignore
├──READMED.md
├──LICENSE
├──......
└─packages
       │ 
       ├── flutter_svg
       │         ├── example
       │         ├── lib
       │         ├── test
       │         ├── README.md
       │         ├── ....
       │         └── pubspec.yaml
       │ 
       └── flutter_svg_test
                   ├── example
                   ├── lib
                   ├── test
                   ├── README.md
                   ├── ....
                   └── pubspec.yaml
```
I am willing to make this work, but you probably need to guide me  a bit.

I can add some more finder like the following one, so the package is not that small.
```
/// Finds widgets created by [SvgPicture.asset] with the [path] argument.
///
/// ## Sample code
/// ```dart
/// expect(svgAssetWithPath('assets/asset_name.svg'), findsOneWidget);
/// ```
/// This will match one [SvgPicture.asset] with the [path] 'assets/asset_name.svg'.
///
/// If the `skipOffstage` argument is true (the default), then this skips
/// nodes that are [Offstage] or that are from inactive [Route]s.
Finder svgAssetWithPath(String path, {bool skipOffstage = true}) {
  return _SvgAssetWithPathFinder(svgPath: path, skipOffstage: skipOffstage);
}

class _SvgAssetWithPathFinder extends MatchFinder {
  _SvgAssetWithPathFinder({required String svgPath, super.skipOffstage})
      : _svgPath = svgPath;

  final String _svgPath;

  @override
  String get description => "Path: '$_svgPath' and created by SvgPicture.asset";

  @override
  bool matches(Element candidate) {
    var result = false;
    final widget = candidate.widget;
    if (widget is SvgPicture) {
      final bytesLoader = widget.bytesLoader;
      if (bytesLoader is SvgAssetLoader) {
        result = bytesLoader.assetName == _svgPath;
      }
    }
    return result;
  }
}
```

dnfield: Yes, this needs to be its own package, which the repo isn't currently structured well for. I think your proposed structure is fine.

packages/flutter_svg_test/README.md
dnfield, line 1: Please update this file with a description of what this package is and how to use it.

packages/flutter_svg_test/flutter_svg_test.iml
dnfield, line 1: I think this should be removed?

packages/flutter_svg_test/lib/svg_test.dart
dnfield, line 1: I don't think this file is necessary

packages/flutter_svg_test/test/flutter_logo.svg
dnfield, line 1: I think this is duplicated from the other package and isn't referred to anywhere, can it be removed?

dnfield:
This looks close. It'll need a conflict resolution. Thanks!

packages/flutter_svg_test/test/flutter_logo.svg
SimonWeidemann: I use this file in two  SvgFileLoader tests.

SimonWeidemann:
I worked through the review comments.
Now the repository has a  readme and each package .  
I thought this would be  more suitable.

packages/flutter_svg_test/test/flutter_svg_test_test.dart
```diff
@@ -0,0 +1,183 @@
...
+
+        expect(find.svg(asset.bytesLoader), findsOneWidget);
+      });
+
+      testWidgets('network svg', (WidgetTester widgetTester) async {
+        await HttpOverrides.runZoned(() async {
+          final SvgPicture asset = SvgPicture.network('svg.dart');
+          await widgetTester.pumpWidget(asset);
+
+          expect(find.svg(asset.bytesLoader), findsOneWidget);
```
dnfield, line 31: So looking at this, it surprises me that we're not just passing in the `SvgPicture`, i.e. `find.svg(asset)`. 

Sorry I didn't see this until now. This is looking pretty close though.

SimonWeidemann: If adjust the finder accordingly:
      
```
        Finder svg(SvgPicture svg, {bool skipOffstage = true}) {
           return _SvgFinder(svg, skipOffstage: skipOffstage);
         }

      class _SvgFinder extends MatchFinder {
        _SvgFinder(this._svg, {super.skipOffstage});
      
         final SvgPicture _svg;
      
         @override
         String get description => "svg: '$_svg'";
      
         @override
         bool matches(Element candidate) {
           return candidate.widget == _svg;
         }
       }
```
For me  the following test should be valid:
```
      testWidgets('asset svg', (WidgetTester widgetTester) async {
        final SvgPicture asset = SvgPicture.asset(
          'test/flutter_logo.svg',
          width: 100,
          height: 400,
        );
        final SvgPicture asset2 = SvgPicture.asset(
          'test/flutter_logo.svg',
          width: 100,
          height: 400,
        );
        await widgetTester.pumpWidget(
          DefaultAssetBundle(
            bundle: _FakeAssetBundle(),
            child: asset,
          ),
        );
```
But unfortunately the test is invalid, because the equals method compares the object not the fields.
If I do this by hand my finder would look like this: 

```
        Finder svg(SvgPicture svg, {bool skipOffstage = true}) {
           return _SvgFinder(svg, skipOffstage: skipOffstage);
         }

      class _SvgFinder extends MatchFinder {
        _SvgFinder(this._svg, {super.skipOffstage});
      
         final SvgPicture _svg;
      
         @override
         String get description => "svg: '$_svg'";
      
         @override
         bool matches(Element candidate) {
           return  widget is SvgPicture &&
                      _svg.runtimeType == widget.runtimeType &&
                     _svg.width == widget.width &&
                     _svg.height == widget.height &&
                     _svg.fit == widget.fit &&
                     _svg.alignment == widget.alignment &&
                     _svg.bytesLoader == widget.bytesLoader &&
                     _svg.placeholderBuilder == widget.placeholderBuilder &&
                     _svg.matchTextDirection == widget.matchTextDirection &&
                     _svg.allowDrawingOutsideViewBox == widget.allowDrawingOutsideViewBox &&
                     _svg.semanticsLabel == widget.semanticsLabel &&
                     _svg.excludeFromSemantics == widget.excludeFromSemantics &&
                     _svg.clipBehavior == widget.clipBehavior &&
                     _svg.colorFilter == widget.colorFilter &&
                     _svg.theme == widget.theme;
         }
       }
```
While writing the  tests helper I orientated on the  `find.text`  and `find.image` from the flutter library.
In both they compare just with a subset of the fields. In `find.text`  with the string and  in`find.image` 
with the image provider. 
I think this was done because otherwise you would have to declare all fields in the test.
Sometimes it is just needed  to locate a widget and afterwards compare different aspects, with  `expect`.
Also sometimes it would be really hard to compare objects during an animation, where for example the double value has a huge amount of decimal place.

I would go with the current implementation.

dnfield: Ahh interesting. I would have assumed the impelementation would be something like `return widget is SvgPicture && _svg.bytesLoader = widget.bytesLoader`, but perhaps this makes more sense. If that's how this works for `Image` though I think it makes sense.

Issue #888, opened 2023-03-16
Error


kirill-21:
```dart
[2023-03-16 07:52:48] [DART] [Build: 3.0.11.0] FormatException: Missing extension byte (at offset 2); 
#0      _Utf8Decoder.convertSingle (dart:convert-patch/convert_patch.dart:1747)
#1      Utf8Decoder.convert (dart:convert/utf.dart:351)
#2      Utf8Codec.decode (dart:convert/utf.dart:63)
#3      SvgAssetLoader.provideSvg (package:flutter_svg/src/loaders.dart:309)
#4      SvgLoader._load.<anonymous closure>.<anonymous closure> (package:flutter_svg/src/loaders.dart:124)
#5      compute.<anonymous closure> (package:flutter/src/foundation/_isolates_io.dart:18)
#6      _RemoteRunner._run (dart:isolate:954)
#7      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:300)
#8      _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192)
```

Have no idea how to reproduce it, this issue is just reported to the application's server with PlatformDispatcher.instance.onError method. Happens to some of users of my app.
![image](https://user-images.githubusercontent.com/63228361/225581414-e82975f6-1c78-43da-bb96-666595b31f2c.png)

PR #889, opened 2023-03-16
Allow malformed UTF-8

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/888


lib/src/loaders.dart
```diff
@@ -37,7 +37,8 @@ class SvgTheme {
   /// See: https://www.w3.org/TR/SVG11/coords.html#Units, https://en.wikipedia.org/wiki/X-height
   final double xHeight;
 
-  vg.SvgTheme _toVgTheme() {
+  /// Creates a [vg.SvgTheme] from this.
+  vg.SvgTheme toVgTheme() {
```
dnfield, line 41: This is separate but needed/intentional.

Issue #865, opened 2023-02-15
currentColor is ignored after 1st load


krokyze:
It was working as expected on `1.1.6` but since updating to `2.0.1` currentColor is ignored after 1st load.

| 1.1.6 | 2.0.1 |
| ------------- | ------------- |
| ![Screenshot_20230216-004257](https://user-images.githubusercontent.com/5682176/219206036-abda7585-3621-4bb5-b678-313c5b6b4794.png) | ![Screenshot_20230216-004103](https://user-images.githubusercontent.com/5682176/219206032-3d54642a-1791-4de9-a5cf-bb912d159c0d.png) |

Example SVG:
```
<svg width="16" height="16" fill="none"
    xmlns="http://www.w3.org/2000/svg">
    <path clip-rule="evenodd" d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14z" stroke="#000" stroke-width="1.2" stroke-linecap="round" stroke-linejoin="round"/>
    <path d="m5 8 2 2 4-4" stroke="currentColor" stroke-width="1.2" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

Example code:
```
Row(
  children: [
    SvgPicture.asset(
      'circle.svg',
      theme: SvgTheme(
        currentColor: Colors.red,
      ),
    ),
    SvgPicture.asset(
      'circle.svg',
      theme: SvgTheme(
        currentColor: Colors.blue,
      ),
    ),
  ],
)
```

virtualzeta:
Hi, I tried your code with a small modification (just so I don't have to save the SVG) and it seems to work fine and I get a blue and a red flag on first load using svgPicture.string instead svgPicture.asset.

Version 2.0.1.

```
Row(
          children: [
            SvgPicture.string(
              circleSvg,
              theme: SvgTheme(
                currentColor: Colors.red,
              ),
            ),
            SvgPicture.string(
              circleSvg,
              theme: SvgTheme(
                currentColor: Colors.blue,
              ),
            ),
          ],
        )
```
```
const String circleSvg = '''
<svg width="16" height="16" fill="none"
    xmlns="http://www.w3.org/2000/svg">
    <path clip-rule="evenodd" d="M8 15A7 7 0 1 0 8 1a7 7 0 0 0 0 14z" stroke="#000" stroke-width="1.2" stroke-linecap="round" stroke-linejoin="round"/>
[...]

PR #652, opened 2022-01-18
Avoid cache invalidation when currentColor/em/rem units change but do not affect SVG

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/642

PR #890, opened 2023-03-16
Make cache respect SvgTheme

dnfield:
Fixes https://github.com/dnfield/flutter_svg/issues/865

See also https://github.com/dnfield/flutter_svg/pull/652

I think it's ok for now because `currentColor` is unlikely to change, and font size is no longer pulled from the build context. I do not intend to make font size come back from context again unless someone really really wants it.


jonahwilliams (approved):
LGTM

PR #896, opened 2023-03-28
Added extensions to prevent issue by handling nulls

### LINKED ISSUE #893: Why the ! operator on message in provideSvg?
Hi!

I created my own `SvgNetworkLoader` extending it. In overriden `prepareMessage` I return `null` on some network failures, as the return type was `Future<Uint8List?>`.

But looks like returning null is illegal here, because in `loaders.dart:361` the library uses `!` on it:

```dart
  String provideSvg(Uint8List? message) => utf8.decode(message!);
```

Which results in an unhandled exception. I propose to either change return type of `prepareMessage` or handle nulls there.

Marco4763:
Implemented null safety extension to Fix #893

### DIFF
--- a/lib/src/loaders.dart
+++ b/lib/src/loaders.dart
@@ -3,6 +3,7 @@ import 'dart:convert' show utf8;
 import 'package:flutter/foundation.dart' hide compute;
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/src/utilities/extensions.dart';
 import 'package:flutter_svg/src/utilities/http.dart';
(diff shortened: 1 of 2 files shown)


dnfield (changes requested):
I don't think this is the right solution. What's really needed is a way to report errors back in the widget, similar to [`Image.errorBuilder`](https://api.flutter.dev/flutter/widgets/Image/errorBuilder.html).

Marco4763:
@dnfield realated to this error report, could you please check the PR in the [vector package pr](https://github.com/dnfield/vector_graphics/pull/196) that is about this feature?

Issue #953, opened 2023-06-26
Is there a replacement for DefaultSvgTheme?


GP4cK:
I'm trying to upgrade to 2.0.7 but DefaultSvgTheme was removed.
Do I need to pass the SvgTheme every time I use a `SvgPicture`? I just want to override the currentColor...

dnfield:
No, this inadvertantly got dropped. Please consider opening a PR to add it back.

PR #964, opened 2023-07-12
Adds DefaultSvgTheme

GP4cK:
Closes https://github.com/dnfield/flutter_svg/issues/953

Some notes/questions:
1. In this PR, I basically copied over the `DefaultSvgTheme` from v1.1.6
2. For the tests, I also copied them except for [this one test](https://github.com/dnfield/flutter_svg/blob/3753ff989c1f7c6ef4781d0772b4a3fe0dde9345/test/default_theme_test.dart#L114-L138) about the `fontSize` being based on the `DefaultTextStyle` because of [this comment](https://github.com/dnfield/flutter_svg/blob/d7b5c23a79dcb5425548879bdb79a5e7f5097ce5/packages/flutter_svg/lib/src/loaders.dart#L20-L26)
3. I noticed that the `SvgTheme` in `flutter_svg/lib/src/loaders.dart` seems to duplicate the one from `vector_graphics_compiler`. Shouldn't we use only the one from `vector_graphics_compiler`?

### DIFF
--- a/packages/flutter_svg/lib/src/loaders.dart
+++ b/packages/flutter_svg/lib/src/loaders.dart
@@ -3,13 +3,14 @@ import 'dart:convert' show utf8;
 import 'package:flutter/foundation.dart' hide compute;
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
-import 'package:flutter_svg/src/utilities/http.dart';
 import 'package:vector_graphics/vector_graphics.dart';
 import 'package:vector_graphics_compiler/vector_graphics_compiler.dart' as vg;
 
 import '../svg.dart' show svg;
+import 'default_theme.dart';
 import 'utilities/compute.dart';
 import 'utilities/file.dart';
+import 'utilities/http.dart';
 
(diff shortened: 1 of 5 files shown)


packages/flutter_svg/lib/src/loaders.dart
```diff
@@ -3,13 +3,14 @@ import 'dart:convert' show utf8;
 import 'package:flutter/foundation.dart' hide compute;
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
-import 'package:flutter_svg/src/utilities/http.dart';
```
GP4cK, line 6: I read somewhere that it's not good to mix relative and absolute imports so I changed it to be more consistent.

dnfield:
> 3. I noticed that the `SvgTheme` in `flutter_svg/lib/src/loaders.dart` seems to duplicate the one from `vector_graphics_compiler`. Shouldn't we use only the one from `vector_graphics_compiler`?

That gets done to avoid exporting types from vector_graphics_compiler to end users from this library.

dnfield (approved):

PR #1000, opened 2023-10-17
Fixes issue where a incorrect key was being given to the ImageCache

bramp:
As discussed in https://github.com/dnfield/flutter_svg/issues/990

In `FlutterVectorGraphicsListener.onImage`, it uses a image cache, and takes the cache key like so: 

```dart
  int _createImageKey(int imageId, int format) {
    return Object.hash(_id, imageId, format);
  }
```

The `_id` is the `loader.hashCode`, which didn't change as the SVG changed, because it looked like this:

```dart
class SvgStringLoader extends SvgLoader<void> {
  /// See class doc.
  const SvgStringLoader(
    this._svg, {
    super.theme,
    super.colorMapper,
  });

  final String _svg;

  @override
  int get hashCode => Object.hash(svg, theme, colorMapper);
```

where it should have been:

```dart
  @override
  int get hashCode => Object.hash(_svg, theme, colorMapper);
```

Such a subtle typo (the missing underscore in front of svg). A similar typo existed in BytesLoader as well.


bramp:
This is currently failing, due to bad formatting (from previous pull request). I'm fixing that here: https://github.com/dnfield/flutter_svg/pull/1001 and I'll rebase when/if #1001 is accepted.

bramp:
Rebased the `dart format` fix, but now `flutter_svg/test/no_width_height_test.dart: Spacing without width or height (failed)` seems to be failing, unrelated to my change :/

dnfield (approved):

PR #1001, opened 2023-10-17
Fixed formatting broken in #999

bramp:
I originally didn't format the debugFillProperties because that seems to be what happens in the main flutter library (presumably for better readability), but here it breaks the presubmits. So here is a quick `dart format`.


dnfield:
Golden test can't be impacted by this change - thanks. I had assumed you weren't working on master and sometimes dartfmt changes slightly :)

PR #1017, opened 2023-12-05
Avoid expanding the parameter type of operator==

srawlins:
Object.operator== takes an `Object`, and no Dart runtime passes a `null` value to an `==` implementation. SvgTheme's implementation should not have an expanded parameter type of `dynamic`.

See https://github.com/flutter/flutter/issues/117838 for the overarching issue for Flutter.


srawlins:
Analysis error in CI looks unrelated.

dnfield:
This looks good, should pass CI after this merge, thanks!

PR #1041, opened 2024-02-20
Relax http

dnfield:


dgilperez:
+1, this prevents the upgrade when http 1.2.1 is not an option for other dependency compatibility issues

PR #1048, opened 2024-03-20
fix(cache): remove entry from _pending in case of exception

IgorKhramtsov:
### Issue
If an exception is thrown during `loader` execution - future with error result will be stuck in the `_pending` map and it will be returned every time `putIfAbsent` is called. There is no way to evict it either, since `evict` method only affects `_cache` map.
The exception in loader may be thrown in case of network error when `SvgNetworkLoader` is used.
### Change
This change adds a `catchError` handler to remove future from the `_pending` map in case of failure. Also a regression test is added that will fail without this fix.


zanderso:

Issue #856, opened 2023-02-12
color in SvgPicture


mahdiqdi:
i used color property to change svg color.but now its deprected.how should i change my svg color

jannisnikoy:
@mahdiqdi the docs seem to suggest using the `colorFilter` property instead. Wish color: remained, even if under the hood it'd use colorFilter. Feel like this is adding a lot of unnecessary code, especially when using lots of svgs 🤷

mahdiqdi:
right now i am using lots of svgs as icon in my flutter app.does colorFilter decrease my performance or not?
@jannisnikoy

felipecastrosales:
Yes, docs suggest [`colorFilter`](https://api.flutter.dev/flutter/dart-ui/BlendMode.html). But I think there could be a better `Migration Guide`.

dnfield:
I would be happy to review any PRs that update the docs and/or read me to help people out with this.

Part of what I'm trying to encourage users of this API to think about is what kind of color filter they want, or whether they want to try to use the `ColorMapper` API instead.

dnfield:
The property is deprecated but still may be used. There is an easy path to migration forward to a more flexible API.

hasanmhallak:
Thanks for the reply, I know it can still be used, but deprecation means eventually this property will be removed.

I'm just saying this is adding a lot of unnecessary code, especially when using lots of svgs. Not all developers are using the colorFilter or need to add  a matrix based filters.

dnfield:
Deprecated means there's a better option. I am in no rush to remove this and am willing to keep it around for a long time to let people migrate :)

petro-i:
Why is the development of this package makes life more difficult? 
Color parameter was so easy to use, but now we have to add one more import at least to have the same functionality. I think color parameter deprecation was an error. Use case of simple "color" parameter is a lot more useful then complex "colorFilter" which is not easy to use and understand how it works in case we neeed just paint svg with single color.

felipecastrosales:
@petro-i, if you need to use only one color, use it like this:

```dart
colorFilter: ColorFilter.mode(YOURCOLOR, BlendMode.srcIn)
```

Unfortunately that's it. But I understand the maintainer's explanation.

HappyMakadiyaS:
Why color is not nullable type in ColorFilter.mode? 
[...]

PR #1065, opened 2024-05-22
Add solidColor parameter to handle colorFilter internally 

da-nish:
### Issue
Regarding this issue: https://github.com/dnfield/flutter_svg/issues/856

### Change
we should keep both like the developer can pass a single color and color filter, internally we handle a check to pick color or color filter. I have added the `solidColor` parameter to handle single color.


zanderso:

PR #1066, opened 2024-05-22
modern sizeOf

domscript:
Prefer using this function over getting the attribute directly from the [MediaQueryData] returned from [of], because using this function will only rebuild the context when this specific attribute changes, not when any attribute changes.


domscript:
Great package by the way

zanderso:

